PR #18, opened 2024-12-12
better parser for step call

TonioSnowden:
Works best for :
```python
from notte.env import NotteEnv
import litellm

goal = "Find best flights from Boston to Tokyo on google flights"
messages = [{"role": "system", "content": goal}]
max_steps = 10

def llm_agent(messages):
    response = litellm.completion(
        model="cerebras/llama3.1-70b",
        messages=messages,
    )
    return response.choices[0].message.content

async def main(max_steps=10):
    async with NotteEnv(headless=False) as env:
        while '<done>' not in messages[-1]['content'] and max_steps > 0:
            resp = llm_agent(messages) # query your llm policy.
            obs = await env.chat(resp) # query notte with llm response.
            messages.append({"role": "assistant", "content": resp})
            messages.append({"role": "user", "content": obs})
            max_steps -= 1

await main()
```

### DIFF
--- a/notte/common/parser.py
+++ b/notte/common/parser.py
@@ -18,7 +18,6 @@ class EnvStepParams(BaseModel):
     action_id: str
     params: dict[str, str] | None
 
-
 class Parser(ABC):
 
     @abstractmethod
@@ -86,8 +85,8 @@ class BaseNotteParser(Parser):
         if action is None or not isinstance(action, str):
             raise ValueError("No action found")
         d = json.loads(action)
-        action_id = list(d.keys())[0]
-        params = d[action_id] if d[action_id] != {} else None
+        action_id = d["action-id"]
+        params = d["params"]
         return EnvStepParams(action_id=action_id, params=params)
 
     @override
@@ -111,8 +110,8 @@ If the action is parameterized, provide the value for each parameter.
 Use the exact following format:
 <action>
 {
-"action-id": {} # if an action has no params
-"action-id": { "name":"value" } # for action with params
+"action-id": "<YOUR_ACTION_ID>",
+"params": { "param_name":"param_value" } 
 }
 </action>
 \nIf you're done, just say <done/>. Nothing else!
--- a/notte/common/parser.py
+++ b/notte/common/parser.py
@@ -18,6 +18,7 @@ class EnvStepParams(BaseModel):
     action_id: str
     params: dict[str, str] | None
 
+
 class Parser(ABC):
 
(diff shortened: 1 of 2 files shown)


notte/common/parser.py
```diff
@@ -111,8 +110,8 @@ def textify(self, obs: Observation) -> str:
 Use the exact following format:
 <action>
 {
-"action-id": {} # if an action has no params
-"action-id": { "name":"value" } # for action with params
+"action-id": "<YOUR_ACTION_ID>",
+"params": { "param_name":"param_value" }
```
giordano-lucas, line 114: probably better to write "<YOUR_PARAM_NAME>": "YOUR_PARAM_VALUE"

notte/common/parser.py
```diff
@@ -86,8 +85,8 @@ def step(self, text: str) -> EnvStepParams:
         if action is None or not isinstance(action, str):
             raise ValueError("No action found")
         d = json.loads(action)
-        action_id = list(d.keys())[0]
-        params = d[action_id] if d[action_id] != {} else None
+        action_id = d["action-id"]
```
giordano-lucas, line 88: @TonioSnowden can you raise a value error if these keys are not present ?

notte/env.py
```diff
@@ -147,7 +147,7 @@ async def chat(self, text: str) -> str:
             return self._parser.textify(obs)
         elif endpoint == "step":
             step_params = self._parser.step(text)
-            obs = await self.step(step_params.action_id, step_params.params, enter=False)
+            obs = await self.step(step_params.action_id, step_params.params, enter=True)
```
giordano-lucas, line 150: I don't think that this is needed

giordano-lucas (changes requested):

andreakiro:
+Precommits are failing, install and fix them before we merge
```
pre-commit install
```

andreakiro (changes requested):
+Precommits are failing, install and fix them before we merge
```
pre-commit install
```

notte/env.py
TonioSnowden: For me needed otherwise it doesn't succeed to insert something (like confirm to put the departure city)

giordano-lucas: I think it's none by default ```
async def step(
        self,
        action_id: str,
        params: dict[str, str] | str | None = None,
        enter: bool | None = None,
    )
   ```

andreakiro (approved):

Issue #21, opened 2024-12-12
Messages above the 8192 llama context length


andreakiro:
despite integrating llamux #17; sometimes len(prompt+document) > 8192 and raises an error

andreakiro:
Use divide and conquer algorithm on context subgraphs

andreakiro:
Closing as not relevant anymore

PR #38, opened 2024-12-14
[wip] greedy-divide-ctx-document

andreakiro:
#21 Still need to;
- [ ] Create a valid pipe for partitioning `pipe(node: NotteNode, gamma: int) -> list[NotteNode]` for generic usage
- [ ] Add a bunch of tests to make sure we cover some of the base edge cases :)

### DIFF
--- a/notte/browser/context.py
+++ b/notte/browser/context.py
@@ -16,8 +16,15 @@ class Context:
     def markdown_description(self) -> str:
         return self.format(self.node, indent_level=0)
 
-    def format(self, node: NotteNode, indent_level: int = 0) -> str:
+    def format(
+        self,
+        node: NotteNode,
+        indent_level: int = 0,  # indentation level for the current node.
+        cumulative_chars: int = 0,  # carries on num_chars from parent nodes.
+        parent_path: list[int] | None = None,  # computes the path to a given node.
+    ) -> str:
         indent = "  " * indent_level
+        parent_path = parent_path or []
 
         # Start with role and optional text
         result = f"{indent}{node.get_role_str()}"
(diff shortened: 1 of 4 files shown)


notte/pipe/main.py
```diff
@@ -46,3 +48,51 @@ def forward(
...
+                partials = [partial_tree_by_path(context.node, path) for path in _paths]
+                merged = merge_trees(partials)
+                if merged is not None:
+                    trees.append(merged)
+            else:
+                path = paths[_partition[0]]
+                partial = partial_tree_by_path(context.node, path)
+                trees.append(partial)
+
+        # _forward for each tree, with incremental listing.
```
giordano-lucas, line 86: I think that it's better to put lines `60-86` in a function `build_context_partitions`  to increase readability / lisibility

notte/browser/context.py
```diff
@@ -16,8 +16,15 @@ def interaction_nodes(self) -> list[InteractionNode]:
     def markdown_description(self) -> str:
         return self.format(self.node, indent_level=0)
 
-    def format(self, node: NotteNode, indent_level: int = 0) -> str:
+    def format(
+        self,
+        node: NotteNode,
+        indent_level: int = 0,  # indentation level for the current node.
```
giordano-lucas, line 22: I think these should probably be computed when the Notte graph is created intead of when it is formated in markdown. In my other PR I created a new function`NotteNode.from_a11y_node` and you can use that to build the necessary variables.

notte/browser/node_type.py
```diff
@@ -405,10 +407,13 @@ class NotteNode:
     id: str | None
     role: NodeRole | str
     text: str
-    subtree_ids: list[str] = field(init=False, default_factory=list)
     children: list["NotteNode"] = field(default_factory=list)
     attributes_pre: NodeAttributesPre = field(default_factory=NodeAttributesPre)
     attributes_post: NotteAttributesPost | None = None
+    subtree_ids: list[str] = field(init=False, default_factory=list)
+    _subtree_chars: int = -1  # length of formatted text for this node+children
```
giordano-lucas, line 414: Why not use `None` for default value ?

notte/pipe/main.py
andreakiro: Yes and it will also be helpful for you in data scraping 👍🏽

notte/browser/context.py
andreakiro: Hmm yes and no. For `_path` I agree, all other variables are strictly correlated with the `format` function as the number of chars depends on the way you build your resulting output string. There is no way for me to predict this ahead of time (before calling `format()` I mean)

notte/browser/node_type.py
andreakiro: Because it is annoying to have to deal with based-pyright type checks when something can be of two different types. Here at least it's always an integer, and my condition is that that thing is positive, that's it.

notte/browser/context.py
giordano-lucas: yeah that's right actually

andreakiro:
#56 up to date

PR #44, opened 2024-12-21
update for api & better agent workflow

giordano-lucas:


notte/pipe/listing.py
```diff
@@ -29,6 +29,11 @@ def forward_incremental(
         context: Context,
         previous_action_list: list[Action],
     ) -> list[PossibleAction]:
+        """
+        This method is used to get the next action list based on the previous action list.
+
+        /!\\ This was designed to only be used in the `forward` method when the previous action list is not empty.
```
andreakiro, line 35:
```suggestion
        //!\\ This was designed to only be used in the `forward` method when the previous action list is not empty.
```

notte/pipe/main.py
```diff
@@ -32,11 +32,8 @@ def forward(
         # we keep only intersection of current context inodes and previous actions!
         previous_action_list = [action for action in previous_action_list if action.id in inodes_ids]
         action_list = self.action_listing_pipe.forward(context, previous_action_list)
-        validated_action = ActionListValidationPipe.forward(inodes_ids, action_list)
 
-        # we merge newly validated actions with the misses we got from previous actions!
-        valided_action_ids = [action.id for action in validated_action]
-        actions: list[Action] = validated_action + [a for a in previous_action_list if a.id not in valided_action_ids]
+        actions = self.merge_action_lists(inodes_ids, action_list, previous_action_list)
```
andreakiro, line 36: Again, I am really NOT a fan of exporting this into a separate function—as this can be a bit misleading. This code is indeed doing merging but e2e the "merging" as we intend it on multiple rounds with previous actions is not only handled by this piece of code. You need also the line 33 to make it work. So it's sort of having a function called `merging` that is only responsible for part of the merging.

andreakiro (approved):

notte/pipe/main.py
giordano-lucas: There is no merging pipeline in the code anymore so I don't think this causes any confusion. Here merging is really used in it's literal sense, i.e. combining 2 lists into a single one 

> `You need also the line 33 to make it work`

This is actually not true, I added the condition `and (a.id in inodes_ids)`. This condition makes the whole processing more explicit and readible in my opinion exactly because it's not dependent of line 33 and can be used and tested externally. 

The reason for adding this function is that,  I realized when I was working on the caching, I add to run some tests to debug that. I needed to check how the merging was operating without having to call the action listing pipeline. 

Since I needed that I might create a function for that. Technically now that this is done, we could remove that function but now that I look at it, I think it improves the readability of the code (we undernstand better that they are different stages in this pipe: (1) listing, (2) merging, (3) check completion threshold). And also if I needed that, I could also need that again in the future so I would vote for keeping it there.

notte/pipe/listing.py
```diff
@@ -65,6 +70,8 @@ def forward(
         context: Context,
         previous_action_list: list[Action] | None = None,
     ) -> list[PossibleAction]:
+        if previous_action_list is not None and len(previous_action_list) > 0:
```
giordano-lucas, line 88: c'était un bug que j'ai solve ça si jamais @andreakiro le forward incremental était plus jamais call

notte/pipe/main.py
giordano-lucas: update: to make the caching work perfectly it was a bit more tricky than I initially expected so that's why I added new function in this class (cf monorepo for more info)

PR #56, opened 2025-01-14
[WIP] greedy-divide-ctx-document

andreakiro:
This PR is pending as it will worsen performance while we don't fix preprocessing issues.

More;
- A few tests need to be added to make sure we don't break stuff in the future
- The context partitioner could be abstracted in a class or module, but not urgent


giordano-lucas:
@andreakiro closing this PR to clean up current PRs. We will re-open that later if needed

PR #73, opened 2025-01-27
Handle links that open new browser tabs.

giordano-lucas:
This PR adds some handler to detect when I new page is created and automatically sets the main page as the newly created page. 

A downside of this approach is that we have to wait an extra `200ms` for the wait event on the new page


giordano-lucas:
already tackled by @leo-notte in #77

PR #80, opened 2025-01-30
Authentication vault

TonioSnowden:


examples/agent.py
```diff
@@ -155,6 +157,12 @@ async def run(self, task: str, url: str | None = None) -> AgentOutput:
         3. When and how to determine task completion
         4. Error handling and recovery strategies
         """
+        credentials, password = get_credentials_for_domain(clean_url(url) if url else None)
+        if credentials and password:
+            logger.info(f"🔑 Using credentials for {url}: {credentials} / {password}")
```
giordano-lucas, line 162: I don't we don't want the credentials to be logged. You can just say `Found credentials for URL` or something like that

examples/credentials.py
```diff
@@ -0,0 +1,28 @@
...
+    password: str
+
+
+def get_credentials_for_domain(url: str | None) -> tuple[str | None, str | None]:
+    """Get credentials for a given domain from environment variables."""
+    if not url:
+        return None, None
+
+    import os
+    from urllib.parse import urlparse
```
giordano-lucas, line 16: It's a bad practice in python to use imports inside function body. Move them on top of file (or even better use the utils file in `utils/url.py`) to parse the url

examples/agent.py
giordano-lucas: Actually I think the logging should only be `Using credential vault` and only when a `login action` is indeed needed then you can have a logging message in the env

examples/credentials.py
```diff
@@ -0,0 +1,28 @@
+import json
+from typing import TypedDict
+
+
+class Credentials(TypedDict):
```
giordano-lucas, line 5: There is a also a `Credentials` dataclass defined below. Is this needed ? Can you at least move them in a single file for consistency ?

giordano-lucas (changes requested):

giordano-lucas:
I don't think I like the proposed approach api. I think it should be better to add an extra actument to the `BaseAgent` that would be the vault instead of loading it implicitely from the `notte.env`. Also why did you use a `singelton` pattern for the vault ? 
> 
> vault = Vault() vault.add_vault(json.load(open("vault_path")))
> 
> agent = SimpleNotteAgent( model="openai/gpt-4o", max_steps=10, headless=False )
> 
> task = "Post 'Hello World' on Twitter" result = await agent.run( task=task, url="https://x.com" # Credentials will be loaded automatically )

giordano-lucas:
For this POC, it would be good also to be able to fetch cookies from a browser (chrome, safari, etc.) and directly fill the vault from that

giordano-lucas:
Can you rename the main folder from `password` to something more explict like `credentials` / `secret_vault` or something like that ?

giordano-lucas:
@TonioSnowden pre-release is out (i.e #77 ) if you want to rebase on main

notte/common/credentials/models.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@dataclass
+class Credentials:
+    url: str
+    username: str
+    password: str
+
+
+class VaultInterface(Protocol):
```
giordano-lucas, line 14: why protocol instead of ABC abstract class ?

giordano-lucas: In the package we use the convention `base.py` for bases classes in a directory (with Base in the class name as well, here it can be BaseVault`. This should be renamed

notte/common/credentials/prompt.py
```diff
@@ -0,0 +1,9 @@
+from pathlib import Path
+
+
+def get_credentials_prompt() -> str:
```
giordano-lucas, line 4: `prompt.py` and `system.md` are too short to justify sperate files. Can you simply add a `def instructions(self) -> str`  or `def system` in `BaseVault` directely ?

pyproject.toml
```diff
@@ -20,6 +20,7 @@ litellm = "^1.59.8"
 chevron = "^0.14.0"
 llamux = "^0.1.9"
 pillow = "^11.1.0"
+hvac = "^2.3.0"
```
giordano-lucas, line 23: this should be an optional dependency I think. Not all users would want this to be installed by default

examples/pipistrello/agent.py
```diff
@@ -0,0 +1,125 @@
...
+from notte.llms.engine import LLMEngine
+
+_ = load_dotenv()
+
+
+class PipistrelloAgentConfig(AgentConfig):
+    env: NotteEnvConfig = NotteEnvConfig().use_llm()
+
+
+class PipistrelloAgent(BaseAgent):
```
giordano-lucas, line 28: now that you finalised the implement I don't think this `Pipistrello` folder is needed anymore. Can you delete that and integrate it directly in falco/gufo ?

giordano-lucas (changes requested):

giordano-lucas (approved):

examples/falco/agent.py
```diff
@@ -102,6 +105,8 @@ def output(self, answer: str, success: bool) -> AgentOutput:
     def get_messages(self, task: str) -> list[AllMessageValues]:
         self.conv.reset()
         system_msg, task_msg = self.prompt.system(), self.prompt.task(task)
+        if self.vault is not None:
+            system_msg += "\n" + self.vault.instructions()
```
giordano-lucas, line 109: cc @leo-notte I think we could think of a tool register component so that this register can automatically append the custom instructions for each of the register tools instead of doing it manually like here. But for later, just something to think about

giordano-lucas (approved):

notte/common/credential_vault/base.py
```diff
@@ -0,0 +1,102 @@
...
+    email_placeholder: ClassVar[str] = "my_email@example.com"
+    password_placeholder: ClassVar[str] = "my_password_123"
+
+    @abstractmethod
+    def add_credentials(self, url: str, username: str, password: str) -> None:
+        """Store credentials for a given URL"""
+        pass
+
+    @abstractmethod
+    def get_credentials(self, url: str) -> VaultCredentials | None:
```
leo-notte, line 29: feel like we would probably want to return a list of VaultCredentials, might need to store multiple users on a single website

notte/common/credential_vault/base.py
```diff
@@ -0,0 +1,102 @@
...
+class VaultCredentials(BaseModel):
+    url: str
+    username: str
+    password: str
+
+
+class BaseVault(ABC):
+    """Base class for vault implementations that handle credential storage and retrieval."""
+
+    email_placeholder: ClassVar[str] = "my_email@example.com"
```
leo-notte, line 20: we might want the placeholder to be a bit more unique, just so we don't unknowingly replace text that isnt specifically a vault pattern to replace?

leo-notte (changes requested):
haven't tested it but looks good with some minor nitpicks

notte/common/credential_vault/base.py
giordano-lucas: Yes I agree but that's probably too complex to handle for now. Let's just wait to see if this feature takes off and when users ask for this we can develop it. At the moment I think its ok to disable multi credentials for the same website

notte/common/credential_vault/base.py
giordano-lucas: Yes good point. Also maybe in the feature we could we also could also create a special `SigninAction` that the llm can selector to be 100% sure that we only do this for sign actions. This should be a special tool / agent that is only responsible for that.

andreakiro:
Are we ready to merge?

PR #82, opened 2025-02-03
replace playwright by patchright

leo-notte:


.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
     hooks:
       - id: mypy
         language: system
-        entry: mypy
```
giordano-lucas, line 25: not sure about that, I wanted the CI to be independent of poetry there should be another way to fix your problem without having to update the CI

giordano-lucas:
Can you rerun the precomit with the correct formating to avoid unnecessary changes ? Otherwise it will be a pain for me to rebase in the huge refactoring pipeline

giordano-lucas (approved):

PR #104, opened 2025-02-11
[temporary?] Disable websec / interactions outside viewport

leo-notte:
https://github.com/nottelabs/notte/pull/93 might be a bit finnicky for now. In the meantime would test with disabled web security so we don't have cors errors

Only difference for simple agent would be:

```python
agent = SimpleAgent(model="cerebras/llama-3.3-70b", headless=False, raise_on_failure=False, disable_web_security=True)
```


giordano-lucas (approved):

notte/browser/pool.py
```diff
@@ -173,22 +173,32 @@ async def _create_browser(self, headless: bool) -> BrowserWithContexts:
...
+
+        browser_args = [
+            "--disable-dev-shm-usage",
+            "--disable-extensions",
+            "--no-sandbox",
+            "--disable-setuid-sandbox",
+            "--no-zygote",
+            "--mute-audio",
+            f'--js-flags="--max-old-space-size={int(self.config.CONTEXT_MEMORY)}"',
+            "--disable-web-security",
```
giordano-lucas, line 185: this should be disabled

PR #128, opened 2025-02-25
Refactor eval and agents + use ruff & uv

giordano-lucas:


.github/workflows/pypi-release.yml
```diff
@@ -71,7 +71,7 @@ jobs:
     name: >-
       GitHub Release
     needs:
-      - publish-to-pypi
```
andreakiro, line 74: Why did you change this?

.pre-commit-config.yaml
andreakiro, line 1: Ruff > Black?

examples/api.py
```diff
@@ -0,0 +1,25 @@
...
+_ = config.cerebras().dev_mode()
+_ = config.env.disable_web_security().not_headless().cerebras().steps(10)
+
+agent = Agent(config=config)
+
+app = FastAPI()
+
+
+@app.get("/")
+def heallth_check():
```
andreakiro, line 15:
```suggestion
def health_check():
```

examples/api.py
andreakiro, line 1: Is this example necessary in this repo?

pyproject.toml
```diff
@@ -1,58 +1,54 @@
-[tool.poetry]
+[project]
 name = "notte"
-version = "1.2.1"
```
andreakiro, line 3: Version is wrong

pyproject.toml
```diff
@@ -1,58 +1,54 @@
...
-description = "The web browser for LLMs"
-authors = ["Notte Team <hello@notte.cc>"]
+version = "0.1.0"
+description = "The web browser for LLMs agents"
```
andreakiro, line 4: We might have to change the description

pyproject.toml
```diff
@@ -1,58 +1,54 @@
...
 readme = "README.md"
+authors = [
+    { name = "Lucas Giordano", email = "lucas@notte.cc" },
+    { name = "Andrea Pinto", email = "andrea@notte.cc" }
+]
 packages = [
```
andreakiro, line 10: This is something we have to discuss but not a fan of this structure

src/notte/browser/snapshot.py
```diff
@@ -40,14 +41,19 @@ class SnapshotMetadata(BaseModel):
     timestamp: dt.datetime = field(default_factory=dt.datetime.now)
 
 
-@dataclass
```
andreakiro, line 43: Why not a dataclass anymore?

src/notte/common/agent/types.py
```diff
@@ -19,3 +20,13 @@ class AgentResponse(BaseModel):
     messages: list[AllMessageValues] | None = None
     llm_usage: list[LlmUsageDictTracer.LlmUsage]
     duration_in_s: float = -1
+
+    @override
+    def __str__(self) -> str:
+        return (
+            f"AgentResponse(success={self.success}, duration_in_s={round(self.duration_in_s, 2)}, answer={self.answer})"
```
andreakiro, line 27: Would be nice to format `self.answer` a bit better in the future

andreakiro: This is the final agent response post-trajectory right?

src/notte/common/fastapi.py
andreakiro, line 1: Doesnt this belong to monorepo?

src/notte/pipe/action/llm_taging/pipe.py
```diff
@@ -170,14 +171,16 @@ def tagging_context(self, context: ProcessedBrowserSnapshot) -> ProcessedBrowser
     def forward(
         self,
         context: ProcessedBrowserSnapshot,
-        previous_action_list: Sequence[Action] | None,  # type: ignore
+        previous_action_list: Sequence[BaseAction] | None,
         pagination: PaginationParams,
     ) -> ActionSpace:
+        # TODO: handle the typing of this properly later on
```
andreakiro, line 177: Can we do now?

.github/workflows/pypi-release.yml
giordano-lucas: because the pipe has been failing for the past few weeks because of of a circular dependency. Not sure what happened but it was not working

pyproject.toml
giordano-lucas: Yeah I was wandering if it would still make sense to change that back to < 1 as we discussed sometime ago

src/notte/pipe/action/llm_taging/pipe.py
giordano-lucas: no too much effort

src/notte/common/fastapi.py
giordano-lucas: I want to be able to quickly spinup an agent api for testing purposes / demo and I think users might be interested to do that. It's an alternative to the cli command to interact with the agent

src/notte/common/agent/types.py
giordano-lucas: not sure I understand what you mean, can you give me an example of what you want to see ? This seems more than enough for me. I just wanted to override this method to avoid the issue where you have to store the result into a variable in notebooks (otherwise we run into memory issues and we have to restart the kernel)

src/notte/browser/snapshot.py
giordano-lucas: because we are using pydantic since the refactoring and I have been slowly changing all dataclasses into BaseModel for consistency

pyproject.toml
giordano-lucas: can you give more details ?

pyproject.toml
giordano-lucas: yeah I agree but for now its more than fine. We will see in next releases

.pre-commit-config.yaml
giordano-lucas: Faster, less error prone and combines linting and formating

pyproject.toml
```diff
@@ -73,11 +69,15 @@ filterwarnings = [
     "ignore:Valid config keys have changed in V2*:UserWarning"
 ]
 
-[tool.black]
+[tool.ruff]
```
leo-notte, line 72: I think we also need

```
[tool.ruff.lint]
extend-select = ["I"]
```

for import sorting

examples/api.py
giordano-lucas: Why not ? I think it's a good way to showcase how this can be used in practice. and personally I like this better than using the command line

giordano-lucas: Also in near the feature, I will add a dockerfile that packages the agent in a container and this piece of code will be used there.

src/notte/pipe/action/llm_taging/pipe.py
giordano-lucas: It has been there since the refactoring with browseruse prompt and its a bit tricky to solve so I would prefer to handle this later as this is again likely to change when we are going to seperate the perception layer fully from the env

examples/api.py
andreakiro: Spinning up an API to run an example is really not a common workflow and I have never seen this. The API is good for containerized apps to be shipped to customers but not for the opensource version. I would keep the focus on CLI and then re-add this api like launcher in case we really need it. 

Ppl will run
```python
$ python run.py --task $task
```
not 
```python 
$ python api.py
$ curl -X POST "http://localhost:8000/agent/run" -H "Content-Type: application/json" -d '{"task": "$task"}'
```

pyproject.toml
andreakiro: No

pyproject.toml
andreakiro: I meant the structure of the package not the authors my bad.
We should be able to install all—agents included—with `pip install notte`

src/notte/browser/snapshot.py
andreakiro: isnt Pydantic adding weight to the package for barely nothing? just asking

src/notte/common/agent/types.py
andreakiro: I mean like having a post-formatter such that you get a nicely formatted response

src/notte/browser/snapshot.py
giordano-lucas: - note: it has always been a requirement for the sdk types 
- it is also a requirement for the browser use pipe
- pydantic is a pretty lightweight package on its own

pyproject.toml
giordano-lucas: We are able to do that using `pip install notte` with the current `packages`

pyproject.toml
giordano-lucas: I reverted it btw but on my side it still think it would be a good idea to do that given the state of the package

PR #139, opened 2025-02-28
Notification response agent

TonioSnowden:
Set-up notifications : 
Code for testing :

```python
from notte.common.agent.types import AgentResponse
import asyncio
import os

# Create a sample AgentResponse
test_response = AgentResponse(
    answer="This is a test notification from Notte",
    success=True,
    env_trajectory=[],
    agent_trajectory=[],
    llm_usage=[],
    duration_in_s=1.5
)

# Test sending a notification
async def test():
    await notifier.notify(
        task="Test notification task",
        result=test_response
    )

# Run the test
await test()
```

Just have to change the notifier set-up.

```python
from notte.common.notifier.email import EmailConfig, EmailNotifier

config = EmailConfig(
    smtp_server=os.getenv("SMTP_SERVER"),
    smtp_port=int(os.getenv("SMTP_PORT", "587")), 
    sender_email=os.getenv("EMAIL_SENDER"),
    sender_password=os.getenv("EMAIL_PASSWORD"),
    receiver_email=os.getenv("EMAIL_RECEIVER")
)

notifier = EmailNotifier(config)
```

```python
from notte.common.notifier.discord import DiscordConfig, DiscordNotifier

config = DiscordConfig(
        token=os.getenv("DISCORD_TOKEN"),
        channel_id=os.getenv("DISCORD_CHANNEL_ID")  

notifier: DiscordNotifier = DiscordNotifier(config)
```

```python
from notte.common.notifier.slack import SlackConfig, SlackNotifier

config = SlackConfig(
    token=os.getenv("SLACK_BOT_TOKEN"),
    channel_id=os.getenv("SLACK_CHANNEL_ID")
)

slack_notifier = SlackNotifier(config)
```


giordano-lucas:
@TonioSnowden can you also add a Readme.md in `common/notifiers` to explain how to setup this tool ?

src/notte/common/notifier/base.py
```diff
@@ -0,0 +1,27 @@
...
+    """Agent wrapper that sends notifications after task completion."""
+
+    def __init__(self, agent: BaseAgent, notifier: BaseNotifier):
+        self.agent: BaseAgent = agent
+        self.notifier: BaseNotifier = notifier
+
+    async def run(self, task: str, url: str | None = None) -> AgentResponse:
+        """Run the agent and send notification about the result."""
+        result = await self.agent.run(task, url)
+        await self.notifier.notify(task, result)
```
giordano-lucas, line 26: I think you forgot to add the discord bot file

src/notte/common/notifier/mail.py
```diff
@@ -0,0 +1,85 @@
...
+        """Send an email notification about the task result.
+
+        Args:
+            task: The task description
+            result: The agent's response to be sent
+        """
+        await self.email_service.connect()
+        try:
+            subject = "Notte agent response"  # Use the task in the subject
+            await self.email_service.send_email(subject, body=result.answer)
```
giordano-lucas, line 83: Can you write a proper mail instead of just dumping the response ?
I.e Hello, {sender}. Here is the response of the agent xxx that you scheduled at .... for the following task: ...

src/notte/common/notifier/mail.py
```diff
@@ -0,0 +1,85 @@
...
+from email.mime.text import MIMEText
+
+from pydantic import BaseModel
+from typing_extensions import override
+
+from notte.common.agent.types import AgentResponse
+from notte.common.notifier.base import BaseNotifier
+
+
+class EmailConfig(BaseModel):
```
giordano-lucas, line 12: can you add one by default for the Notte Email server ? ask @andreakiro for the credentials. The `sender_password` should be taken from the .env file.

src/notte/common/notifier/slack.py
```diff
@@ -0,0 +1,54 @@
...
+        self._client: WebClient = WebClient(token=self.config.token)
+
+    async def send_message(self, text: str) -> None:
+        """Send a message to the configured Slack channel.
+
+        Args:
+            text: The message text to send
+        """
+        try:
+            _: SlackResponse = self._client.chat_postMessage(channel=self.config.channel_id, text=text)  #  type: ignore[unknown import]
```
giordano-lucas, line 33: Why did you add a `try/catch` block ? And actually it would be usefull to provide more feedback to the user in case of an error.

src/notte/common/notifier/slack.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @override
+    async def notify(self, task: str, result: AgentResponse) -> None:
+        """Send a Slack notification about the task result.
+
+        Args:
+            task: The task description
+            result: The agent's response to be sent
+        """
+        message = f"*Task*: {task}\n*Response*: {result.answer}"
```
giordano-lucas, line 53: same for format of the message here

uv.lock
giordano-lucas, line 1: I think you forgot to add the `pyproject.toml` file

src/notte/common/notifier/base.py
```diff
@@ -0,0 +1,27 @@
...
+
+from notte.common.agent.base import BaseAgent
+from notte.common.agent.types import AgentResponse
+
+
+class BaseNotifier(ABC):
+    """Base class for notification implementations."""
+
+    @abstractmethod
+    async def notify(self, task: str, result: AgentResponse) -> None:
```
giordano-lucas, line 11: why is this async ?

src/notte/common/notifier/base.py
TonioSnowden: yes

src/notte/common/notifier/mail.py
TonioSnowden: Dis moi si tu veux plus parce que sender and schedule at je peux pas les avoir depuis la Agent response and si il y a besoin de plus faut que j'ajoute en de parametres

src/notte/common/notifier/slack.py
TonioSnowden: I thought is was just a good practice because we can have multiple error here if the user doesn't have the correct credentials, or fill a wrong token...
Do I change it to just the call function

src/notte/common/notifier/base.py
TonioSnowden: My intuition is that when you have http request and server connection especially the one with the mail for server connection and the one with discord for when you have to get the channel and the server connection. it's a good practice to have an async function.

src/notte/common/notifier/slack.py
```diff
@@ -0,0 +1,67 @@
...
+        """Send a Slack notification about the task result.
+
+        Args:
+            task: The task description
+            result: The agent's response to be sent
+        """
+        message = f"""
+:robot_face: *Notte Agent Report*
+
+*Task Details*
```
giordano-lucas, line 56: can you put this formating function inside the `BaseNotifier` class to avoid code duplication inside the 3 subclasses ?

pyproject.toml
```diff
@@ -42,6 +42,9 @@ api = [
     "fastapi>=0.115.8",
     "uvicorn>=0.29.0",
 ]
+discord = [
+    "discord-py>=2.5.0",
```
giordano-lucas, line 46: In this list we should have all packages that you import in `common/notifiers` that are not already defined in the dependency section of the pyproject.toml (i.e smtplib, email, slack_sdk). In python, its a convention that if you have an `import X` line in your code. `x` must be present in the pyproject.toml somehow.

src/notte/common/notifier/base.py
giordano-lucas: I am not sure about that. If none of the Smpt server, Slack client or discord client require async function I would remove that `async` to simplify the code

src/notte/common/notifier/discord.py
```diff
@@ -0,0 +1,79 @@
...
+    token: str
+    channel_id: int
+
+
+class DiscordService:
+    """Service for sending messages to Discord from Notte."""
+
+    def __init__(self, config: DiscordConfig):
+        self.config: DiscordConfig = config
+        intents = discord.Intents.default()
```
giordano-lucas, line 22: @TonioSnowden there are still at lot of `check warning on line XX in src/notte/common/notifier/XXX` to solve in case you didn't see

src/notte/common/notifier/slack.py
giordano-lucas: Yes please simply these pieces of code. When you write 
```
try:
            _: SlackResponse = self._client.chat_postMessage(channel=self.config.channel_id, text=text)  #  type: ignore[unknown import]
        except SlackApiError as e:
            raise ValueError(f"Failed to send Slack message: {str(e)}")
```
It does not help the user in any way to `group` all exceptions into a single exception. As a matter of fact, it reduces the amount of information a user has access to be able to solve these issues. E.g. `SlackApiError` is more informative to resolve an error comming from the Slack SDK than a Value Error coming from the notte library.

Also in general, if you are not doing anything to actually fix an error raised by an exception in the catch/finally block, it usually means that the try is useless and should be remove.

giordano-lucas:
@TonioSnowden I don't think the `SlackService` , `DiscordService` add any value to the code. Can you put everything into the corresponding `Notifier` class ? It would make things simpler and clearer for end users (i.e one single class to look at)

pyproject.toml
TonioSnowden: Yes forgot slack_sdk. smtplib and email are part of the python package.

TonioSnowden:
@giordano-lucas the send_message of notifier/mail.py don't really sure if it's the correct way of doing it. Because we can't define it properly because of the email_structure and argument passed

src/notte/common/notifier/mail.py
```diff
@@ -0,0 +1,101 @@
...
+        msg["To"] = self.config.receiver_email
+        msg["Subject"] = subject
+
+        msg.attach(MIMEText(body, "plain"))
+
+        if self._server:
+            _ = self._server.send_message(msg)
+
+    @override
+    async def send_message(self, text: str) -> None:
```
giordano-lucas, line 61: what's the purpose of this function ? I don't see the point of overriding to call the parent function. It does not make any sense to me

giordano-lucas:
> @giordano-lucas the send_message of notifier/mail.py don't really sure if it's the correct way of doing it. Because we can't define it properly because of the email_structure and argument passed

Can you give me an example ? I don't think I understand the issue from your comment

src/notte/common/notifier/base.py
giordano-lucas: can you add a comment in the code to explain that ?

src/notte/common/notifier/slack.py
giordano-lucas: @TonioSnowden still relevant I think

TonioSnowden:
@giordano-lucas basedpyright really don't recognize discord import. Had to type ignore everything

src/notte/common/notifier/base.py
```diff
@@ -0,0 +1,50 @@
...
+
+class BaseNotifier(ABC):
+    """Base class for notification implementations."""
+
+    @abstractmethod
+    async def send_message(self, text: str) -> None:
+        """Send a message using the specific notification service."""
+        pass
+
+    async def notify(self, task: str, result: AgentResponse) -> None:
```
giordano-lucas, line 15: can you add a  `def format_message` function instead of having this inside `notitif` in that way you can override this function in the email client to avoid code duplication (cf next comment)

src/notte/common/notifier/mail.py
```diff
@@ -0,0 +1,97 @@
...
+        await self.connect()
+        try:
+            body = f"""
+Hello,
+
+This is an automated message from your Notte Agent.
+
+Task Details:
+-------------
+Task: {task}
```
giordano-lucas, line 79: *code duplication*. Please tackle this. You should override `def format_message(...) = Hello,
This is an automated message from your Notte Agent. ... super.format_message(...) ... Best regards,
Your Notte Agent 🌒

giordano-lucas (approved):
LGTM

PR #146, opened 2025-03-04
make config immuable

giordano-lucas:
Config are now immuable :

```
config = AgentConfig().cerebras().map_env(lambda env: (env.disable_web_security().not_headless().cerebras().steps(15)))
```

### DIFF
--- a/examples/cli_agent.py
+++ b/examples/cli_agent.py
@@ -1,5 +1,7 @@
 import asyncio
 
+from dotenv import load_dotenv
+
 from notte_agents.falco.agent import (
     FalcoAgent as Agent,
 )
@@ -7,12 +9,14 @@ from notte_agents.falco.agent import (
     FalcoAgentConfig as AgentConfig,
 )
 
+# Load environment variables
+_ = load_dotenv()
+
 if __name__ == "__main__":
     parser = AgentConfig.create_parser()
     _ = parser.add_argument("--task", type=str, required=True, help="The task to run the agent on.")
     args = parser.parse_args()
-    config = AgentConfig.from_args(args)
-    _ = config.env.disable_web_security().user_mode()
+    config = AgentConfig.from_args(args).map_env(lambda env: env.disable_web_security().user_mode())
     agent = Agent(config=config)
 
     out = asyncio.run(agent.run(args.task))
@@ -21,4 +25,4 @@ if __name__ == "__main__":
 
 # uvicorn examples.api:app --reload
 # export task="Go to twitter.com and login with the account lucasgiordano@gmail.com and password mypassword8982"
-# python run.py --task $task
+# python run.py --task $task --reasoning_model "openai/gpt-4o"
--- a/examples/cli_agent.py
+++ b/examples/cli_agent.py
@@ -16,13 +16,12 @@ if __name__ == "__main__":
     parser = AgentConfig.create_parser()
     _ = parser.add_argument("--task", type=str, required=True, help="The task to run the agent on.")
     args = parser.parse_args()
-    config = AgentConfig.from_args(args).map_env(lambda env: env.disable_web_security().user_mode())
+    config = AgentConfig.from_args(args).map_env(lambda env: env.user_mode())
     agent = Agent(config=config)
 
     out = asyncio.run(agent.run(args.task))
     print(out)
 
 
-# uvicorn examples.api:app --reload
 # export task="Go to twitter.com and login with the account lucasgiordano@gmail.com and password mypassword8982"
-# python run.py --task $task --reasoning_model "openai/gpt-4o"
+# python run.py --task $task --reasoning_model "openai/gpt-4o" --env.headless False --env.disable_web_security True
--- a/src/notte/browser/pool/local_pool.py
+++ b/src/notte/browser/pool/local_pool.py
@@ -55,15 +55,18 @@ class MemoryBrowserPoolConfig(FrozenConfig):
 class BrowserPoolConfig(FrozenConfig):
     memory: MemoryBrowserPoolConfig = MemoryBrowserPoolConfig()
     base_debug_port: int = 9222
-    disable_web_security: bool = False
+    web_security: bool = False
     max_browsers: int | None = None
     max_total_contexts: int | None = None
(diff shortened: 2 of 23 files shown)


examples/cli_agent.py
```diff
@@ -1,18 +1,22 @@
...
+# Load environment variables
+_ = load_dotenv()
+
 if __name__ == "__main__":
     parser = AgentConfig.create_parser()
     _ = parser.add_argument("--task", type=str, required=True, help="The task to run the agent on.")
     args = parser.parse_args()
-    config = AgentConfig.from_args(args)
-    _ = config.env.disable_web_security().user_mode()
+    config = AgentConfig.from_args(args).map_env(lambda env: env.disable_web_security().user_mode())
```
leo-notte, line 19: not a fan of disable_web_security being the default still

src/notte/browser/pool/local_pool.py
```diff
@@ -63,29 +66,33 @@ class BrowserPoolConfig(BaseModel):
     viewport_width: int = 1280
     viewport_height: int = 1020  # Default in playright is 720
 
+    def set_disable_web_security(self: T) -> T:
+        return self._copy_and_validate(disable_web_security=True)
+
+    @cached_property
```
leo-notte, line 72: would remove the get_ prefix if we make them properties

src/notte/env.py
```diff
@@ -50,7 +49,10 @@ class ScrapeAndObserveParamsDict(ScrapeParamsDict, PaginationParamsDict):
     pass
 
 
-class NotteEnvConfig(BaseModel):
+TNotteEnvConfig = TypeVar("TNotteEnvConfig", bound="NotteEnvConfig")
```
leo-notte, line 52: could use https://docs.python.org/3/library/typing.html#typing.Self instead, right?

andreakiro:
+imo we can kill the `examples` folder and simply add the `cli_agent` inside the `notte` package as this will become the base endpoint for opensource users

giordano-lucas:
> +imo we can kill the `examples` folder and simply add the `cli_agent` inside the `notte` package as this will become the base endpoint for opensource users

I don't agree with this, I think the cli_agent, vault_agent, api_agent all serve different purposes and each of them deserve to be in the example files

examples/cli_agent.py
giordano-lucas: I agree but since this is in `examples` I thought it was ok. But I can also all a argument in the parser

src/notte/browser/pool/local_pool.py
giordano-lucas: Ilt's just not make them properties then

src/notte/browser/pool/local_pool.py
```diff
@@ -63,17 +62,18 @@ class BrowserPoolConfig(BaseModel):
     viewport_width: int = 1280
     viewport_height: int = 1020  # Default in playright is 720
 
+    def set_disable_web_security(self: Self) -> Self:
```
leo-notte, line 65: for these bool vars it feels a bit weird to have set and disable and the same time, would maybe rather have 
```python
def set_web_security(self, val : bool)
  ...
```
or
```python
def enable_web_security(self, val : bool)
  ...
def disable_web_security(self, val : bool)
  ...
```

src/notte/browser/window.py
```diff
@@ -27,22 +29,35 @@
...
-class BrowserWindowConfig(BaseModel):
+class BrowserWindowConfig(FrozenConfig):
     headless: bool = False
     pool: BrowserPoolConfig = BrowserPoolConfig()
     wait: BrowserWaitConfig = BrowserWaitConfig()
     screenshot: bool | None = True
     empty_page_max_retry: int = 5
     cdp_url: str | None = None
 
+    def set_headless(self: Self, value: bool | None = None) -> Self:
```
leo-notte, line 48: didn't ask before but since I'm being nitpicky, might as well ask now, why do we default the value to None instead of the actual default, if we just default to True in the function itself?

src/notte/common/agent/config.py
```diff
@@ -40,32 +42,49 @@ class AgentConfig(BaseModel):
...
+
+    def cerebras(self: Self) -> Self:
+        return self._copy_and_validate(reasoning_model="cerebras/llama-3.3-70b")
+
+    def use_vision(self: Self) -> Self:
+        return self._copy_and_validate(include_screenshot=True)
+
+    def dev_mode(self: Self) -> Self:
+        return self._copy_and_validate(
+            raise_condition=RaiseCondition.IMMEDIATELY,
```
leo-notte, line 80: I had this change on my other branch, but I think it makes sense to pull RaiseCondition out of dev_mode, and to just make dev mode be about verbose output.

Then also allow changing the raise condition if we actually want to raise.
```python
def set_raise_condition(self, value: RaiseCondition)
```

src/notte/browser/pool/local_pool.py
leo-notte: Could also make it lru_cache if we really cared about it but yea fine by me

leo-notte (approved):

src/notte/browser/pool/local_pool.py
giordano-lucas: Cannot have disable_web_security because it's already a an argument in the pydantic model. But I can add a value like we have in set_headless

src/notte/common/agent/config.py
leo-notte: I mean I see that we can achieve it by doing this but I don't see why dev_mode would have a different meaning depending on if it's applied to the agent vs the env

src/notte/browser/pool/local_pool.py
giordano-lucas: I remaned the argument `disable_web_security` into `web_security` and then I could implement your change

src/notte/common/agent/config.py
giordano-lucas: I think you point, it's just that I wanted to have a quick way to enable all login + see the exception stack trace right away when I am debugging (that's what I called agent dev mode)

PR #147, opened 2025-03-05
144 external step callback for agents status

giordano-lucas:
Now we can define a callback in gufo and falco that is called inside the step function (usefull for logging purposes


leo-notte (approved):
not the biggest fan of writing the traces in the root directory / creating folders in site packages for people that pip install the package, but otherwise looks fine to me

Issue #143, opened 2025-03-04
`notte_agents` and `notte_eval` are not available in wheel file


giordano-lucas:
Currently the only way to create an environment with `notte` as a dependency and access the `notte_agents` package is to do the following:

```
git clone https://github.com/nottelabs/notte
uv pip install -e notte
```

I.e

```
un run python
> importe notte_eval
```

We need to fix this so that `uv build` actually creates a wheel file that contains the `notte_eval` and `notte_agents` code. Or even better, let users install
```
pip install notte[agent, eval]
```
so that they can only install the packages that they are interested in

giordano-lucas:
@andreakiro it would be good time to also update the `release cicd` so that it uses `uv build`

PR #148, opened 2025-03-06
move agents inside notte for dist

andreakiro:
Moved agents zoo inside the `notte` package to simplify distro. This is a quick fix for #143.
We'll add better packaging for evals and integrations in the future.


giordano-lucas (approved):
all good after cicd fixes

PR #149, opened 2025-03-06
Fix cli config

giordano-lucas:
* this should run : ```uv run examples/cli_agent.py --task $task --reasoning_model "openai/gpt-4o" --env.disable_web_security```
* added a test to avoid getting out of sync in the future


andreakiro (approved):

andreakiro:
Ruff fails for the cicd

PR #152, opened 2025-03-06
Collection of falco fixes

leo-notte:
Main takeaways:
- Retry on llm structured output validation fail
  - Add some specific validation errors that hint to the llm how to fix
  - Similar for scraping: add hints to ensure that `short_data` always had data instead of None
  - Ensure that at least one action is always selected by the llm at validation time
- Add RelevantInteractions to llm output in falco: meant to ensure that only actions on existing / relevant interaction ids are being taken
  - This probably needs to be revamped to make sure we are more clear on what is an interaction vs an action
- More conservative timeouts (could probably be tuned down by now, most changes don't affect that much)
- Allow accessibility tree to be None (since everything should be fine with just dom)
- Add params for anchor to select proxy / captcha  
- Remove base64 images from simple md scraping
- Conversation autosize: switch to using litellm for token count (also takes into account images), set to True by default on falco
- Ensure scraping prompts have both a system and user message (otherwise gemini simply errors out)

In hindsight most of those changes should've been atomic commits, each linked to an issue :man_shrugging:


src/notte/common/tools/conversation.py
```diff
@@ -43,26 +44,27 @@ class Conversation:
     json_extractor: StructuredContent = field(default_factory=lambda: StructuredContent(inner_tag="json"))
     autosize: bool = False
     max_tokens: int = 16000
-    encoding: tiktoken.Encoding = tiktoken.get_encoding("cl100k_base")
+    model: str = "openai/gpt-4o"
+    conservative_factor: float = 1.0
```
giordano-lucas, line 48: if you think that this will fail most of the time with 1.0, should we just set that by default to 0.8-0.9 and remove this default value in the falco init method ?

leo-notte: I was debating it, if you pick a model that is [supported by litellm](https://docs.litellm.ai/docs/completion/token_usage) it should give you the right amount, but I've had it fail on me with llama, so I kind of agree

src/notte/data/space.py
```diff
@@ -60,6 +60,18 @@ def wrap_dict_in_root_model(cls, values: dict[str, Any]) -> dict[str, Any]:
             values["success"] = False
         return values
 
+    @model_validator(mode="after")
+    def ensure_data_if_success(self) -> Self:
+        if self.success and self.data is None:
+            raise ValueError("Scraping was successful but data field is None")
```
giordano-lucas, line 66: can you create specific exception classes in `notte/errors` for that ?

src/notte/data/space.py
```diff
@@ -60,6 +60,18 @@ def wrap_dict_in_root_model(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+        return self
+
+    @model_validator(mode="after")
+    def ensure_no_error_success(self) -> Self:
+        if self.success and (self.error is not None and self.error != ""):
+            raise ValueError("If error, make sure success is False. If success, make sure error is null.")
```
giordano-lucas, line 72: same here

src/notte/errors/handler.py
```diff
@@ -46,14 +47,22 @@ async def wrapper(*args: Any, **kwargs: Any) -> T:
             logging.error(f"NotteBaseError: {e.dev_message}")
             raise e
         except PlaywrightTimeoutError as e:
-            # logging.error(f"PlaywrightTimeoutError with message: {e.message}")
-            # Transform external timeout error
-            if "- waiting for locator(" in str(e):
+            # only timeout issue if the last line is it
```
giordano-lucas, line 50: Just for me to understand the change here, the issue is that an element way not be loaded so the `InvalidLocatorRuntimeError` was incorrectly raised ?

src/notte/errors/handler.py
```diff
@@ -46,14 +47,22 @@ async def wrapper(*args: Any, **kwargs: Any) -> T:
...
                 raise InvalidLocatorRuntimeError(message=str(e)) from e
+
             raise PlaywrightRuntimeError(message=str(e)) from e
         except TimeoutError as e:
             raise NotteTimeoutError(message="Request timed out.") from e
         # Add more except blocks for other external errors
+        except PlayrightError as e:
+            raise NotteBaseError(
+                dev_message=f"Unexpected playwright error: {str(e)}",
+                user_message="An unexpected playwright error occurred. Our team has been notified.",
```
giordano-lucas, line 62: no need to mention playright in `user_message` (these are for the users of the api that may not have any idea about what playwright is)

src/notte/llms/engine.py
```diff
@@ -50,20 +55,37 @@ def structured_completion(
         messages: list[AllMessageValues],
         response_format: type[TResponseFormat],
         model: str | None = None,
+        retries: int = 3,
```
giordano-lucas, line 58: can this be set to 0 | None by default ? And override this parameter in a config somwhere in `AgentConfig` or `EnvConfig` ?

src/notte/errors/handler.py
leo-notte: the whole string would basically be
```
- waiting for locator
  - waiting to be visible
  - element is actionable and visible
```
but since at one point we waited on locator, we would raise as a timeout when it wasn't always the case

src/notte/pipe/scraping/pipe.py
```diff
@@ -1,6 +1,7 @@
 from enum import StrEnum
 from typing import Self, final
 
+from html2text import config
```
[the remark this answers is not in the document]
leo-notte: we could if we really want to make sure it's there but it's the main dependency from main_content_extractor so it would likely not change

src/notte/agents/falco/agent.py
```diff
@@ -170,11 +187,25 @@ async def step(self, task: str) -> CompletionAction | None:
...
+
             self.trajectory.add_step(result)
             step_msg = self.trajectory.perceive_step_result(result, include_ids=True)
             if not result.success:
                 logger.error(f"🚨 {step_msg}")
+
+                # observe again
+                obs = await self.env.observe()
+                ex_status = ExecutionStatus(
+                    input=typing.cast(BaseAction, FallbackObserveAction()),
```
giordano-lucas, line 201: why do we need the type casting ? I though FallbackObserveAction was a subclass of BaseAction)

src/notte/llms/prompts/extract-json-schema/multi-entity/user.md
```diff
@@ -0,0 +1 @@
+Transformed content:
```
giordano-lucas, line 1: Suggestion, maybe just adding a action / verb would yield better results. Your turn, provide me with the transformed content" or something like that

src/notte/agents/falco/agent.py
```diff
@@ -151,14 +160,22 @@ def get_messages(self, task: str) -> list[AllMessageValues]:
         if last_valid_obs is not None and self.history_type is not HistoryType.FULL_CONVERSATION:
             self.conv.add_user_message(
                 content=self.perception.perceive(last_valid_obs),
-                image=last_valid_obs.screenshot if self.config.include_screenshot else None,
+                image=(last_valid_obs.screenshot if self.config.include_screenshot else None),
+            )
+
+        if len(self.trajectory.steps) > 0:
+            self.conv.add_user_message(
+                content="""Given the previous information, start by reflecting on your last action. Then, summarize the current page and list relevant available interactions.
```
giordano-lucas, line 168: can you move these instuctions into the promp.py object directly ?

giordano-lucas (approved):
I left you a few comments but LGTM

src/notte/controller/base.py
```diff
@@ -100,10 +100,10 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
         match action:
             # Interaction actions
             case ClickAction():
-                await locator.click()
+                await locator.click(timeout=5000)
```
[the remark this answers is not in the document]
giordano-lucas: yes or just static functions like `BrowserWaitConfig.short_wait` or `...long_wait`

src/notte/errors/handler.py
giordano-lucas: But overall it was still a TimeoutError raised by Playright, right ? So do you simply think it was better to keep the raw exception which contains more information if we later on feed that to the Agent in case of an error ?

leo-notte: yeah my issue was that the error gets turned into an InvalidLocatorRuntimeError instead of a TimeoutError which I don't think is right

The error would then have
            agent_message=(
                "Execution failed because interactive element is not found or not visible. "
                "Hint: wait 5s and try again, check for any modal/dialog/popup that might be blocking the element,"
                " or try another action."
 
When in actuality, the element was found and visible within the timeout window

giordano-lucas: but then what would be your "advice/recommendation" to the LLM in that case this happens ? Should we simply provide a better error message instead of the raw playright output ?

leo-notte: For now just the raw playwright output is likely more useful than the blanket timeout message, but yeah ideally would add specific messages for each log that playwright has, unfortunately I didn't find the source of those messages yet

PR #157, opened 2025-03-07
[FillAction] Text containing indentation should be pasted instead of filled char by char (e.g. code, bullet points, etc)

TonioSnowden:
## Summary by CodeRabbit

- **New Features**
  - Added a platform-specific control key mapping to ensure the correct modifier key is used based on the operating system.
  - Introduced an option to automatically clear text during fill actions.
  - Simplified vault initialization process for credential management.
  - Enhanced clipboard access permissions for browser contexts.
  - Introduced a function to detect significant indentation in text.
  - Added asynchronous tests to verify clipboard isolation between different browser contexts.

- **Refactor**
  - Streamlined authentication setup by simplifying vault initialization from environment variables.
  - Updated text handling logic to focus on detecting significant indentation for improved formatting and interaction.

- **Bug Fixes**
  - Improved error handling for credential management.

- **Tests**
  - Added test cases to validate the behavior of the new text indentation detection function.
  - Introduced tests for clipboard isolation functionality.

### DIFF
--- a/examples/leetcode_agent.py
+++ b/examples/leetcode_agent.py
@@ -1,4 +1,3 @@
-import asyncio
 import os
 
 from dotenv import load_dotenv
@@ -8,30 +7,25 @@ from notte.common.agent.types import AgentResponse
 from notte.common.credential_vault.hashicorp.vault import HashiCorpVault
 
 # Load environment variables and create vault
+# Required environment variables for HashiCorp Vault:
+# - VAULT_URL: The URL of your HashiCorp Vault server
+# - VAULT_DEV_ROOT_TOKEN_ID: The root token for authentication in dev mode
 _ = load_dotenv()
 vault = HashiCorpVault.create_from_env()
 
 # Add leetcode credentials
-leetcode_username = os.getenv("LEETCODE_USERNAME")
-leetcode_password = os.getenv("LEETCODE_PASSWORD")
-if not leetcode_username or not leetcode_password:
-    raise ValueError("LEETCODE_USERNAME and LEETCODE_PASSWORD must be set in the .env file.")
-
-vault.add_credentials(url="https://leetcode.com", username=leetcode_username, password=leetcode_password)
+vault.add_credentials(
+    url="https://leetcode.com", username=os.getenv("LEETCODE_USERNAME"), password=os.getenv("LEETCODE_PASSWORD")
+)
 
 agent: Agent = Agent(vault=vault)
 
-
-async def main() -> AgentResponse:
-    return agent.run(
+if __name__ == "__main__":
+    response: AgentResponse = agent.run(
         task=(
             "Go to leetcode.com and solve the problem of the day. when you arrive on the page change the programming language to python."
             "First login to leetcode and then resolve the problem of the day"
             "When there is a cloudflare challenge, click on the box to verify that you are human"
         )
     )
-
-
-if __name__ == "__main__":
-    response: AgentResponse = asyncio.run(main())
     print(response)
--- a/examples/leetcode_agent.py
+++ b/examples/leetcode_agent.py
@@ -3,65 +3,35 @@ import os
 
 from dotenv import load_dotenv
 
-from notte.agents.falco.agent import FalcoAgent as Agent
-from notte.agents.falco.agent import FalcoAgentConfig as AgentConfig
+from notte.agents import Agent
+from notte.common.agent.types import AgentResponse
 from notte.common.credential_vault.hashicorp.vault import HashiCorpVault
-from notte_integrations.remote_sessions.steel_pool import SteelBrowserPool
 
-# Load environment variables
+# Load environment variables and create vault
 _ = load_dotenv()
+vault = HashiCorpVault.create_from_env()
 
-vault_url = os.getenv("VAULT_URL")
-vault_token = os.getenv("VAULT_DEV_ROOT_TOKEN_ID")
-if not vault_url or not vault_token:
-    raise ValueError(""""
-VAULT_URL and VAULT_DEV_ROOT_TOKEN_ID must be set in the .env file.
-For example if you are running the vault locally:
-
-```
-VAULT_URL=http://0.0.0.0:8200
-VAULT_DEV_ROOT_TOKEN_ID=<your-vault-dev-root-token-id>
-```
-
-""")
-
-try:
-    vault = HashiCorpVault(url=vault_url, token=vault_token)
-except ConnectionError:
-    vault_not_running_instructions = """
-Make sure to start the vault server before running the agent.
-Instructions to start the vault server:
-> cd src/notte/common/credential_vault/hashicorp
-> docker-compose --env-file ../../../../../.env up
-"""
-    raise ValueError(f"Vault server is not running. {vault_not_running_instructions}")
-
+# Add leetcode credentials
 leetcode_username = os.getenv("LEETCODE_USERNAME")
 leetcode_password = os.getenv("LEETCODE_PASSWORD")
-
 if not leetcode_username or not leetcode_password:
     raise ValueError("LEETCODE_USERNAME and LEETCODE_PASSWORD must be set in the .env file.")
 
 vault.add_credentials(url="https://leetcode.com", username=leetcode_username, password=leetcode_password)
 
-steel_pool = SteelBrowserPool(verbose=True)
+agent: Agent = Agent(vault=vault)
 
 
-async def main():
-    await steel_pool.start()
-
-    config = AgentConfig().openai().map_env(lambda env: (env.disable_web_security().not_headless().steps(15)))
-    agent = Agent(config=config, vault=vault, pool=steel_pool)
-
-    output = await agent.run(
-        (
+async def main() -> AgentResponse:
+    return agent.run(
+        task=(
             "Go to leetcode.com and solve the problem of the day. when you arrive on the page change the programming language to python."
             "First login to leetcode and then resolve the problem of the day"
             "When there is a cloudflare challenge, click on the box to verify that you are human"
         )
     )
-    print(output)
 
 
-# Run the async function
-asyncio.run(main())
+if __name__ == "__main__":
+    response: AgentResponse = asyncio.run(main())
+    print(response)
--- a/examples/leetcode_agent.py
+++ b/examples/leetcode_agent.py
@@ -1,0 +1,67 @@
+import asyncio
+import os
+
+from dotenv import load_dotenv
+
+from notte.agents.falco.agent import FalcoAgent as Agent
+from notte.agents.falco.agent import FalcoAgentConfig as AgentConfig
+from notte.common.credential_vault.hashicorp.vault import HashiCorpVault
+from notte_integrations.remote_sessions.steel_pool import SteelBrowserPool
+
+# Load environment variables
+_ = load_dotenv()
+
+vault_url = os.getenv("VAULT_URL")
+vault_token = os.getenv("VAULT_DEV_ROOT_TOKEN_ID")
+if not vault_url or not vault_token:
+    raise ValueError(""""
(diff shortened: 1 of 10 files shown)


src/notte/controller/base.py
```diff
@@ -94,18 +99,43 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
         if action.press_enter is not None:
             press_enter = action.press_enter
         # locate element (possibly in iframe)
-        locator = await locale_element(self.window.page, action.selector)
+        locator: Locator = await locale_element(self.window.page, action.selector)
         original_url = self.window.page.url
 
         action_timeout = self.window.config.wait.action_timeout
 
+        def is_code(text: str) -> bool:
```
giordano-lucas, line 107: can you move that into a new `notte/utils/code.py` file ? Can you also add some unit tests ? I want to make sure that we are not detecting too much code. Unit tests should contain at least 3 full positive examples (i.e only code from 3 different languages). Then 3 full negative examples (i.e empty string, "hello word" and some long text, at least 3 paragraphs). Then at least 3 mixed example (natural text + code mixed together)

giordano-lucas: Rationale: the controller should not be responsible of `detecting code`. The controller only contain code for `executing actions in playwright context`

src/notte/controller/base.py
```diff
@@ -94,18 +99,43 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
...
+                return False
+
         match action:
             # Interaction actions
             case ClickAction():
                 await locator.click(timeout=action_timeout)
             case FillAction(value=value):
-                await locator.fill(value, timeout=action_timeout)
-                await self.window.page.wait_for_timeout(500)
+                if is_code(text=value):
```
giordano-lucas, line 119: can add a logging message: "Code detected in fill action: simulate clipboard copy/paste for better string formating" or smeting like that

src/notte/controller/base.py
```diff
@@ -94,18 +99,43 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
...
+                    await locator.focus()
+                    selectallKey: Literal["Meta+A", "Control+A"] = (
+                        "Meta+A" if platform.system() == "Darwin" else "Control+A"
+                    )
+                    await self.window.page.keyboard.press(key=selectallKey)
```
giordano-lucas, line 124: I would suggest addting `await self.window.short_wait()` in between all `press` just to be sure

src/notte/controller/base.py
```diff
@@ -94,18 +99,43 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
...
+                    await self.window.page.keyboard.press(key="Backspace")
+
+                    await self.window.page.evaluate("text => {navigator.clipboard.writeText(text)}", value)
+
+                    # Paste using keyboard shortcut
+                    pasteKey: Literal["Meta+V", "Control+V"] = (
+                        "Meta+V" if platform.system() == "Darwin" else "Control+V"
```
giordano-lucas, line 131: create a function `platform_control_key() -> Litteral["Darwin" | "Control"]` and remove all variables  `pasteKey', `copykey`, etc => to put directly `...press(key=f"{platform_control_key()+V"}` , etc.

src/notte/controller/base.py
```diff
@@ -94,18 +99,43 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
...
         match action:
             # Interaction actions
             case ClickAction():
                 await locator.click(timeout=action_timeout)
             case FillAction(value=value):
-                await locator.fill(value, timeout=action_timeout)
-                await self.window.page.wait_for_timeout(500)
+                if is_code(text=value):
+                    await locator.focus()
+                    selectallKey: Literal["Meta+A", "Control+A"] = (
```
giordano-lucas, line 121: Add a parameter in `FillAction` for that like `clear_text: bool = True`. Then, the agent would be automatically able to `clear_text = False` if it does not want to clear the text.

examples/leetcode_agent.py
```diff
@@ -0,0 +1,67 @@
...
+
+leetcode_username = os.getenv("LEETCODE_USERNAME")
+leetcode_password = os.getenv("LEETCODE_PASSWORD")
+
+if not leetcode_username or not leetcode_password:
+    raise ValueError("LEETCODE_USERNAME and LEETCODE_PASSWORD must be set in the .env file.")
+
+vault.add_credentials(url="https://leetcode.com", username=leetcode_username, password=leetcode_password)
+
+steel_pool = SteelBrowserPool(verbose=True)
```
giordano-lucas, line 47: don't add that directly in the example. We wan't users to use our system instead

examples/leetcode_agent.py
```diff
@@ -0,0 +1,67 @@
+import asyncio
+import os
+
+from dotenv import load_dotenv
+
+from notte.agents.falco.agent import FalcoAgent as Agent
```
giordano-lucas, line 6: can you refactor using `from notte.agents import Agent` ?

examples/leetcode_agent.py
```diff
@@ -0,0 +1,67 @@
...
+from notte.agents.falco.agent import FalcoAgentConfig as AgentConfig
+from notte.common.credential_vault.hashicorp.vault import HashiCorpVault
+from notte_integrations.remote_sessions.steel_pool import SteelBrowserPool
+
+# Load environment variables
+_ = load_dotenv()
+
+vault_url = os.getenv("VAULT_URL")
```
giordano-lucas, line 14: I think there some code duplication with `vault_agent.py` now. Can you fix that ? Also I am thinking, could we automatically load `VAULT_URL and `VAULT_DEV_ROOT_TOKEN_ID` in `HashiCorpVault` it would make this simpler for users to use the vault

giordano-lucas (changes requested):
I left you some comments @TonioSnowden

examples/vault_agent.py
```diff
@@ -2,51 +2,23 @@
...
 
 from dotenv import load_dotenv
-from requests.exceptions import ConnectionError
 
 from notte.agents.falco.agent import FalcoAgent as Agent
 from notte.agents.falco.agent import FalcoAgentConfig as AgentConfig
 from notte.common.credential_vault.hashicorp.vault import HashiCorpVault
 
-# Load environment variables
+# Load environment variables and create vault
```
giordano-lucas, line 10: add a comment to explain to users that they have to setup:
VAULT_URL
VAULT_DEV_ROOT_TOKEN_ID in the .env

src/notte/common/credential_vault/hashicorp/vault.py
```diff
@@ -1,7 +1,9 @@
+import os
 from dataclasses import dataclass
 from typing import Any, Protocol, final
 from urllib.parse import urlparse
 
+from dotenv import load_dotenv
```
giordano-lucas, line 6: That should not be there. This line should be moved back to `vault_agents.py`. It's the user's responsability to load their env file (in whatever way they want to).

src/notte/common/credential_vault/hashicorp/vault.py
```diff
@@ -92,3 +94,40 @@ def get_credentials(self, url: str) -> VaultCredentials | None:
...
+        try:
+            return cls(url=vault_url, token=vault_token)
+        except ConnectionError:
+            vault_not_running_instructions = """
+Make sure to start the vault server before running the agent.
+Instructions to start the vault server:
+> cd src/notte/common/credential_vault/hashicorp
+> docker-compose --env-file ../../../../../.env up
+"""
+            raise ValueError(f"Vault server is not running. {vault_not_running_instructions}")
```
giordano-lucas, line 133: can you keep the exception chain intact by `raising ValueError(..) from e` where `except ConnectionError as e`

src/notte/controller/actions.py
```diff
@@ -288,6 +288,7 @@ class FillAction(InteractionAction):
     id: str
     description: str = "Fill an input field with a value"
     value: str
+    clear_text: bool = True
```
giordano-lucas, line 291: I don't think this parameter is explicit enough for the LLM to understand what this does. I though more about something like `clear_before_fill`

src/notte/controller/base.py
```diff
@@ -1,4 +1,5 @@
 from loguru import logger
+from patchright.async_api._generated import Locator
```
giordano-lucas, line 2:
```suggestion
from patchright.async_api import Locator
```

src/notte/utils/code.py
```diff
@@ -0,0 +1,22 @@
+def has_indentation(text: str) -> bool:
```
giordano-lucas, line 1: a clearer name would be `text_contains_tabs` indentation could mean different things. We

src/notte/controller/base.py
```diff
@@ -104,8 +107,28 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
...
+                        await self.window.short_wait()
+
+                    await self.window.page.evaluate("text => {navigator.clipboard.writeText(text)}", arg=value)
+
+                    # Paste using keyboard shortcut
+                    await self.window.page.keyboard.press(key=f"{platform_control_key()}+V")
+                    await self.window.short_wait()
+                else:
+                    await locator.fill(value, timeout=action_timeout, force=action.clear_text)
+                    await self.window.page.wait_for_timeout(timeout=500)
```
giordano-lucas, line 131:
```suggestion
                    await self.window.short_wait()
```

examples/leetcode_agent.py
```diff
@@ -0,0 +1,37 @@
...
+from notte.agents import Agent
+from notte.common.agent.types import AgentResponse
+from notte.common.credential_vault.hashicorp.vault import HashiCorpVault
+
+# Load environment variables and create vault
+_ = load_dotenv()
+vault = HashiCorpVault.create_from_env()
+
+# Add leetcode credentials
+leetcode_username = os.getenv("LEETCODE_USERNAME")
```
giordano-lucas, line 15: same comment as in `vault_agent`

examples/leetcode_agent.py
```diff
@@ -0,0 +1,37 @@
...
+        task=(
+            "Go to leetcode.com and solve the problem of the day. when you arrive on the page change the programming language to python."
+            "First login to leetcode and then resolve the problem of the day"
+            "When there is a cloudflare challenge, click on the box to verify that you are human"
+        )
+    )
+
+
+if __name__ == "__main__":
+    response: AgentResponse = asyncio.run(main())
```
giordano-lucas, line 36:
```suggestion
    response: AgentResponse = agent.run(
        task=...)
```

with `Agent` now you can simply use `.run` withot asyncio. This can be simplified.

src/notte/controller/base.py
```diff
@@ -104,8 +107,28 @@ async def execute_interaction_action(self, action: InteractionAction) -> Browser
...
+                    await locator.focus()
+
+                    if action.clear_before_fill:
+                        # Select all text
+                        await self.window.page.keyboard.press(key=f"{platform_control_key()}+A")
+                        await self.window.short_wait()
+                        await self.window.page.keyboard.press(key="Backspace")
+                        await self.window.short_wait()
+
+                    await self.window.page.evaluate("text => {navigator.clipboard.writeText(text)}", arg=value)
```
leo-notte, line 124: might be nitpicky, but I think this could create leaks in when hosting multiple browsers on the same machine, as I imagine they still share the clipboard, right?

giordano-lucas: Yes I agree with that @TonioSnowden can you take a look at this ? their might be ways to provide clipboard isolation

giordano-lucas (changes requested):
Figure out clipboard isolation

src/notte/controller/base.py
TonioSnowden: @leo-notte this solution didn't worked it don't keep the structure. I used isolated clipboard which is working

giordano-lucas (approved):

Issue #168, opened 2025-03-12
Include connection to proxy in arguments


leo-notte:
Basically just passing down
```
browser = await chromium.launch(proxy={
  "server": "http://myproxy.com:3128",
  "username": "usr",
  "password": "pwd"
})
```
through NotteEnv

Having residential proxies will be relevant for many people, including the hosted part of notte

PR #170, opened 2025-03-13
Add user_agent and proxy as parameters

### LINKED ISSUE #169: Enforce having user agent if headless is set to True
Using patchright is a decent easy way to bypass some bot detections, but when setting `headless=True`, our user-agent is set to a headless user-agent, which means instant detection. 
Instead of providing a default user-agent, I would just enforce that a user agent has to be provided if `headless=True`.

In the future, would look into browserforge, for getting consistent fingerprints (including usera-gent)

leo-notte:
Fixes #169 #168 
Idea is to be able to just run:

`
uv run python examples/cli_agent.py --task "go to https://ifconfig.co/json and return your ip and user agent" --env.user_agent "my_user_agent" --env.proxy='{"server":"<server_url>","username":"<user>","password":"<pass>"}'        
`
> `AgentResponse(success=True, duration_in_s=4.58, answer=IP address: <proxy ip>, User agent: my_user_agent) `

As mentioned in #169, headless user agent is not great, so we now enforce that headless=True means that we set a user agent manually

### DIFF
--- a/tests/integration/test_user_agent.py
+++ b/tests/integration/test_user_agent.py
@@ -1,0 +1,16 @@
+import json
+
+import pytest
+
+from notte.env import NotteEnv, NotteEnvConfig
+
+
+@pytest.mark.asyncio
+async def test_user_agent():
+    """Test validation of special action parameters"""
+
+    USER_AGENT = "test-user-agent"
+    async with NotteEnv(NotteEnvConfig().set_user_agent(USER_AGENT).headless()) as env:
+        _ = await env.goto("https://ifconfig.co/json")
+        json_resp = json.loads(env.snapshot.dom_node.inner_text())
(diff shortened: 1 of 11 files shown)


tests/integration/test_user_agent.py
```diff
@@ -0,0 +1,16 @@
...
+
+from notte.env import NotteEnv, NotteEnvConfig
+
+
+@pytest.mark.asyncio
+async def test_user_agent():
+    """Test validation of special action parameters"""
+
+    USER_AGENT = "test-user-agent"
+    async with NotteEnv(NotteEnvConfig().set_user_agent(USER_AGENT).headless()) as env:
```
leo-notte, line 13: @giordano-lucas there's a bit of an issue I see in the current handling of configs vs having a builder, and checking once we call build: 
If we set first `headless()`, it fails because `user_agent` hasn't been set yet

PR #208, opened 2025-04-07
Persona SDK

leo-notte:
## Summary by CodeRabbit

- **New Features**
  - Introduced persona management capabilities, allowing the API to now support operations such as creating personas, generating virtual numbers, and retrieving communications like emails and SMS.

- **Refactor**
  - Enhanced the structure and clarity of request and response models to improve overall API consistency.
  - Reformatted type annotations for better readability across various classes.
  - Added a new client to manage persona interactions within the API.

### DIFF
--- a/src/notte/sdk/endpoints/persona.py
+++ b/src/notte/sdk/endpoints/persona.py
@@ -45,7 +45,7 @@ class PersonaClient(BaseClient):
 
         Initializes the client with an optional API key for persona management.
         """
-        super().__init__(base_endpoint_path="persona", api_key=api_key)
+        super().__init__(base_endpoint_path="personas", api_key=api_key)
 
     @override
     @staticmethod
--- a/src/notte/sdk/endpoints/persona.py
+++ b/src/notte/sdk/endpoints/persona.py
@@ -41,20 +41,18 @@ class PersonaClient(BaseClient):
         api_key: str | None = None,
     ):
         """
-        Initialize a SessionsClient instance.
+        Initialize a PersonaClient instance.
 
-        Initializes the client with an optional API key and server URL for session management,
-        setting the base endpoint to "sessions". Also initializes the last session response to None.
+        Initializes the client with an optional API key for persona management.
         """
         super().__init__(base_endpoint_path="persona", api_key=api_key)
 
     @override
     @staticmethod
     def endpoints() -> Sequence[NotteEndpoint[BaseModel]]:
-        """Returns a sequence of available session endpoints.
+        """Returns the available persona endpoints.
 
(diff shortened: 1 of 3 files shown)


src/notte/sdk/endpoints/persona.py
```diff
@@ -0,0 +1,169 @@
...
+    def __init__(
+        self,
+        api_key: str | None = None,
+    ):
+        """
+        Initialize a PersonaClient instance.
+
+        Initializes the client with an optional API key for persona management.
+        """
+        super().__init__(base_endpoint_path="persona", api_key=api_key)
```
giordano-lucas, line 48: would set it to `personas` similarly to other endpoints (`sessions`, `agents`, etc.)

src/notte/sdk/endpoints/persona.py
```diff
@@ -0,0 +1,169 @@
...
+    Client for the Notte API.
+
+    Note: this client is only able to handle one session at a time.
+    If you need to handle multiple sessions, you need to create a new client for each session.
+    """
+
+    # Session
+    EMAILS_READ = "{persona_id}/email/read"
+    SMS_READ = "{persona_id}/sms/read"
+    CREATE_NUMBER = "{persona_id}/create-number"
```
giordano-lucas, line 36: do we need this endpoint to be exposed to endpoint users ? We can keep it for now but I am not sure we want to allow users to create multiple phone number per persona. Also then we should have `{persona_id}/sms/{phone_number_id}/read`

src/notte/sdk/types.py
```diff
@@ -164,6 +175,76 @@ def session_closed() -> "SessionDebugRecordingEvent":
...
+
+class SMSReadRequestDict(TypedDict, total=False):
+    limit: int
+    timedelta: dt.timedelta | None
+    unread_only: bool
+
+
+class SMSReadRequest(BaseModel):
+    limit: int = 10
+    timedelta: dt.timedelta | None = None
```
giordano-lucas, line 213: what's the purpose of timedelta ?

giordano-lucas: messages from now until now - timedelta ?

giordano-lucas: Maybe we can find a more explicit name (or at least write a better description)

src/notte/sdk/endpoints/persona.py
leo-notte: the reason for `create_number` is that by default, a persona doesnt have a phone number. it cant have multiple, at most one. this is because a lot of people might not need one by default, but we might want to add it later one. 
i could add a parameter to create_persona as well if that makes it clearer that we dont have one by default

PR #268, opened 2025-04-25
Adding pdf draft

quangnamta:
Can you have a look at the draft implementation of the pdf features

## Summary by CodeRabbit

- **New Features**
  - Added support for scraping PDF content directly from URLs, extracting and returning content in markdown format.
  - Introduced an example script demonstrating how to scrape PDF pages from arXiv.
- **Dependencies**
  - Added a new dependency on the "docling" package.
- **Tests**
  - Added a new integration test to verify PDF scraping functionality from arXiv.

### DIFF
--- a/.github/workflows/test-cicd.yml
+++ b/.github/workflows/test-cicd.yml
@@ -106,4 +106,4 @@ jobs:
 
       - name: Run unit tests
         run: |
-          uv run pytest tests --ignore=tests/integration/test_resolution.py --ignore=tests/integration/test_webvoyager_resolution.py --ignore=tests/integration/test_e2e.py --ignore=tests/integration/test_webvoyager_scripts.py --ignore=tests/examples/test_examples.py --ignore=tests/examples/test_readme.py --durations=10
+          uv run pytest tests --ignore=tests/integration/test_webvoyager_resolution.py --ignore=tests/integration/test_e2e.py --ignore=tests/examples/test_examples.py --ignore=tests/examples/test_readme.py --durations=10
--- a/packages/notte-browser/src/notte_browser/playwright.py
+++ b/packages/notte-browser/src/notte_browser/playwright.py
@@ -157,6 +157,8 @@ class WindowManager(PlaywrightManager):
                     "width": options.viewport_width,
                     "height": options.viewport_height,
                 }
+            else:
+                logger.warning("No viewport set, using default viewport in playwright")
 
             context: BrowserContext = await self.browser.new_context(
                 # no viewport should be False for headless browsers
--- a/packages/notte-browser/src/notte_browser/window.py
+++ b/packages/notte-browser/src/notte_browser/window.py
@@ -148,7 +148,7 @@ class BrowserWaitConfig(FrozenConfig):
     GOTO: ClassVar[int] = 10_000
     GOTO_RETRY: ClassVar[int] = 1_000
     RETRY: ClassVar[int] = 1_000
-    STEP: ClassVar[int] = 1_000
+    STEP: ClassVar[int] = 5_000
     SHORT_WAIT: ClassVar[int] = 500
     ACTION_TIMEOUT: ClassVar[int] = 1_000
 
--- a/packages/notte-core/src/notte_core/actions/base.py
+++ b/packages/notte-core/src/notte_core/actions/base.py
@@ -58,7 +58,8 @@ class PossibleAction(BaseModel):
             case "M":
                 return "misc"
             case "F":
-                raise NotImplementedError("Image actions are not supported")
+                # figure / image
+                return "image"
             case "S":
                 return "special"
             case _:
--- a/packages/notte-core/src/notte_core/browser/dom_tree.py
+++ b/packages/notte-core/src/notte_core/browser/dom_tree.py
@@ -286,30 +286,6 @@ class DomAttributes:
                 attrs[key] = value
         return attrs
 
-    @staticmethod
-    def from_a11y_node(node: A11yNode) -> "DomAttributes":
-        remaning_keys = set(node.keys()).difference(
-            [
-                "children",
-                "children_roles_count",
-                "nb_pruned_children",
-                "group_role",
-                "group_roles",
-                "markdown",
-                "id",
-                "path",
-                "role",
-                "name",
-                "level",
-                "only_text_roles",
-                # Add any other irrelevant keys here
-                "orientation",
-                "eid",
-                "method",
-            ]
-        )
-        return DomAttributes.safe_init(**{key: node[key] for key in remaning_keys})  # type: ignore
-
     @override
     def __repr__(self) -> str:
         # only display relevant attributes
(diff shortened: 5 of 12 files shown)


giordano-lucas:
@quangnamta thank you for the PR, I will review the code and get back to you

giordano-lucas:
I think I don't like the idea of make the read PDF a separate action in the controller. It really does not have to be. It's a functionally of the scrape action. When the url to scrape is a file (i.e ends with `.pdf` or whatever) we can trigger a parsing call to `doctling`. This has to be hidden to the end user and agent interacting with notte in my opinion

giordano-lucas:
Any reason why you use a dockerfile instead of  `pip install docling` ?

examples/using_read_pdf.py
```diff
@@ -0,0 +1,16 @@
+import asyncio
+
+from notte_browser.session import NotteSession
+
+
+async def read_pdf(des_url: str):
+    async with NotteSession() as page:
+        data = await page.read_pdf(url=des_url)
```
giordano-lucas, line 8: would prefer to have in `page.scrape` actually (cf other comment)

giordano-lucas:
Can you also write some tests in the `tests` folder to validate your implemenation ?

packages/notte-browser/src/notte_browser/controller.py
```diff
@@ -86,6 +87,8 @@ async def execute_browser_action(self, window: BrowserWindow, action: BaseAction
                     await window.page.keyboard.press("PageDown")
             case ScrapeAction():
                 raise NotImplementedError("Scrape action is not supported in the browser controller")
+            case ReadPDFAction():
```
giordano-lucas, line 90: I would remove this action and put into the scraping behaviour (cf other comments)

packages/notte-browser/src/notte_browser/session.py
```diff
@@ -438,6 +441,20 @@ async def scrape(
         self.obs.data = data
         return data
 
+    # Consider putting into god since this might be special action
+    @timeit(name="read_pdf")
+    @track_usage("page.read_pdf")
+    async def read_pdf(
```
giordano-lucas, line 447: cf other comments

packages/notte-integrations/src/notte_integrations/pdf/pdf_options.py
```diff
@@ -0,0 +1,26 @@
...
+
+from notte_integrations.pdf.docling.config import DEFAULT_CONFIG
+from notte_integrations.pdf.docling.serve import DoclingPDFHandler
+
+
+class PDFOptionEnum(StrEnum):
+    DOCLING = "Docling"
+
+
+class PDFChoiceHandler:
```
giordano-lucas, line 12: This can be simplified at this stage. We don't plan to support many pdf tools as this point. You can remove this

packages/notte-browser/src/notte_browser/session.py
```diff
@@ -24,6 +24,8 @@
 from notte_core.llms.engine import LlmModel
 from notte_core.llms.service import LLMService
 from notte_core.utils.webp_replay import ScreenshotReplay, WebpReplay
+from notte_integrations.pdf.docling.serve import DoclingPDFHandler
```
giordano-lucas, line 27: integrations should not be imported in notte. We don't want to create this dependency. You can import `BasePDFHandler` though and that should be enough.

packages/notte-integrations/src/notte_integrations/pdf/docling/config.py
```diff
@@ -0,0 +1,35 @@
+from enum import StrEnum
```
giordano-lucas, line 1: let's put the content in `serve.py` (there is only 30 lines of code`)

packages/notte-integrations/src/notte_integrations/pdf/pdf_options.py
```diff
@@ -0,0 +1,26 @@
...
+    @classmethod
+    def create_from_env(cls):
```
giordano-lucas, line 14: Is the url going to change ? Does it really have to be set in the env file ? I would remove `create_from_env` and just put that in the constructor of the class

examples/using_read_pdf.py
quangnamta: Should I used it as a like plugin integration (modular) or it just write straight to .scrape()?

giordano-lucas:
> Because I believe docling installation (nvidia cuda, etc) is somewhat big and will cloud notte dependency with like nvidia - cuda and things

what do you mean I don't understand what you are saying ? This is the open-source repo, what we are doing on the cloud version does not affect this choice

examples/using_read_pdf.py
quangnamta:
```python
from notte_browser.session import NotteSession

async with NotteSession() as page:
    data = await page.scrape(url="https://www.notte.cc")
    print(data.markdown)
```

Here is the code in the docs. You would prefer to have like page.scrape(url="https://www.notte.cc", pdf=DoclingPDF) right?

packages/notte-integrations/src/notte_integrations/pdf/pdf_options.py
quangnamta: Should i still keep the BasePDFHandler or should I just import docling right into scrape and parse?

packages/notte-browser/pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "notte_core==0.0.dev",
     "patchright==1.50.0",
     "maincontentextractor",
+    "docling==2.31.0"
```
giordano-lucas, line 19: I don't think we want to include `docling` in the notte-browser dependencies. Like you did before in `notte-integrations` was perfect. You can keep the base class you had in the previous version

packages/notte-browser/src/notte_browser/scraping/pdf_scraping.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    PDF scraping pipe that scrapes data from a pdf
+    """
+
+    def __init__(
+        self,
+    ):
+        pass
+
+    async def forward(self, url: str):
```
giordano-lucas, line 15: should go in `notte-integrations`

packages/notte-browser/pyproject.toml
giordano-lucas: You can add a new optional parameter in the NotteSession contructior (pdf_pipe) and that should be good enough for now

packages/notte-browser/src/notte_browser/session.py
```diff
@@ -432,7 +435,16 @@ async def scrape(
         **scrape_params: Unpack[ScrapeParamsDict],
     ) -> DataSpace:
         if url is not None:
+            # Check for application/pdf
+            client = httpx.AsyncClient()
```
giordano-lucas, line 439: I don't think we need to do this check + it will be slow for large PDF documents. I think just checking the `.pdf` part in the url should be enough

packages/notte-browser/src/notte_browser/session.py
```diff
@@ -4,6 +4,7 @@
 from collections.abc import Callable, Sequence
```
giordano-lucas, line 4: also it would be good to automatically trigger this pipe in the `preobserve` function (similar to what we are doing with the `auto_scrape` parameter at the moment. the reason for this is that on a PDF page there is technically no action to take except scrape so we can fastforward this to the agent and already fetch the content.

packages/notte-browser/src/notte_browser/session.py
quangnamta: The reason I do this is because like arxiv there is no explicit .pdf in the url, and on somepage, it's not even exist.

packages/notte-browser/src/notte_browser/session.py
quangnamta: Yes it is true that I check yesterday it is automatically raise error. And if the page is not pdf then docling will raise error. The only reason why I'm doing it this way is to explicitly making it easy to read. Lmk what you think. Plus homepage does not have pdf explicitly

quangnamta: Also there is more content-type other than application/pdf

packages/notte-browser/src/notte_browser/session.py
quangnamta: For ex it can be: notte.cc/file/1234:5678. In which it will completely skip

packages/notte-browser/pyproject.toml
quangnamta: I will have that fix tomorrow

quangnamta:
@giordano-lucas I close the previoius pull request because I have massive uv.sync conflict and my change is not really big so I will make a pull request again on this topics.

PR #294, opened 2025-05-07
Add get cookies, switch sessions

leo-notte:
@giordano-lucas Issue is that if we use non null session, then we overwrite params from agent and it's a bit of a config mess again, any idea on what's best? I can just keep it to using window for now for my example until we change configs

(some tests will fail until api updated) 

## Summary by CodeRabbit

- **New Features**
  - Added the ability to set and retrieve browser session cookies through new asynchronous methods and API endpoints.
  - Introduced support for serializing cookies to JSON files.

- **Refactor**
  - Improved session and cookie management with more explicit and asynchronous handling.
  - Renamed several cookie upload methods and classes for clarity (e.g., "upload_cookies" to "set_cookies").
  - Updated agent session lifecycle management for greater control and reliability.

- **Bug Fixes**
  - Corrected cookie expiration dates in integration tests.

- **Tests**
  - Added new tests for setting and retrieving cookies, and for verifying round-trip cookie operations.

- **Chores**
  - Added "google-auth" as a new dependency.

- **Documentation**
  - Updated example scripts to use explicit asynchronous session management.

### DIFF
--- a/docs/sdk_tutorial.md
+++ b/docs/sdk_tutorial.md
@@ -13,7 +13,7 @@ session = client.sessions.start(timeout_minutes=5)
 # get the session status
 status = client.sessions.status(session.session_id)
 # list your active sessions
-active_sessions = client.sessions.list()
+active_sessions = client.sessions.list_sessions()
 # stop your session
 client.sessions.stop(session.session_id)
 ```
--- a/docs/sdk_tutorial.md
+++ b/docs/sdk_tutorial.md
@@ -13,7 +13,7 @@ session = client.sessions.start(timeout_minutes=5)
 # get the session status
 status = client.sessions.status(session.session_id)
 # list your active sessions
-active_sessions = client.sessions.list_sessions()
+active_sessions = client.sessions.list()
 # stop your session
 client.sessions.stop(session.session_id)
 ```
--- a/packages/notte-agent/src/notte_agent/falco/agent.py
+++ b/packages/notte-agent/src/notte_agent/falco/agent.py
@@ -72,23 +72,13 @@ class FalcoAgent(BaseAgent):
     def __init__(
         self,
         config: FalcoAgentConfig,
-        window: BrowserWindow | None = None,
-        session: NotteSession | None = None,
+        window: BrowserWindow,
         vault: BaseVault | None = None,
         step_callback: Callable[[str, StepAgentOutput], None] | None = None,
     ):
-        if session is not None and window is not None:
-            raise ValueError("Can't set window and session at the same time")
-
-        import logging
-        logging.warning(f"{session=}")
-
-        if session is None:
-            session = NotteSession(config=config.session, window=window)
-
-        logging.warning(f"{session=} {window=}")
-
+        session = NotteSession(config=config.session, window=window)
         super().__init__(session=session)
+
         self.config: FalcoAgentConfig = config
         self.vault: BaseVault | None = vault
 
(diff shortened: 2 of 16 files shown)


docs/sdk_tutorial.md
```diff
@@ -13,7 +13,7 @@ session = client.sessions.start(timeout_minutes=5)
 # get the session status
 status = client.sessions.status(session.session_id)
 # list your active sessions
-active_sessions = client.sessions.list()
+active_sessions = client.sessions.list_sessions()
```
giordano-lucas, line 16: can we keep `.list()` ? I prefer the shorter syntax + it's compatible with steel & browserbase

packages/notte-agent/src/notte_agent/falco/agent.py
```diff
@@ -148,12 +160,12 @@ async def reset(self) -> None:
         self.step_executor.reset()
         await self.session.reset()
 
-    def output(self, answer: str, success: bool) -> AgentResponse:
+    async def output(self, answer: str, success: bool) -> AgentResponse:
```
giordano-lucas, line 163: why do we need async ?

packages/notte-agent/src/notte_agent/falco/agent.py
```diff
@@ -297,8 +309,12 @@ async def _run(self, task: str, url: str | None = None) -> AgentResponse:
         if url is not None:
             task = f"Start on '{url}' and {task}"
 
-        # Loop through the steps
-        async with self.session:
+        started_before_run = self.session.started
```
giordano-lucas, line 312: I don't like this, is this really needed ? I would prefer to raise an exception is the session is not started, especially if the session comes from the sdk. Otherwise the users don't know the sessions have been started and they have no way to know they should close it

packages/notte-agent/src/notte_agent/falco/agent.py
```diff
@@ -314,15 +330,15 @@ async def _run(self, task: str, url: str | None = None) -> AgentResponse:
                 # validate the output
                 if not output.success:
                     logger.error(f"🚨 Agent terminated early with failure: {output.answer}")
-                    return self.output(output.answer, False)
+                    return await self.output(output.answer, False)
```
giordano-lucas, line 333: to be removed

packages/notte-agent/src/notte_agent/falco/agent.py
```diff
@@ -314,15 +330,15 @@ async def _run(self, task: str, url: str | None = None) -> AgentResponse:
...
                 # Sucessful execution and LLM output is not None
                 # Need to validate the output
                 logger.info(f"🔥 Validating agent output:\n{output.model_dump_json()}")
                 val = self.validator.validate(task, output, self.session.trajectory[-1])
                 if val.is_valid:
                     # Successfully validated the output
                     logger.info("✅ Task completed successfully")
-                    return self.output(output.answer, output.success)
+                    return await self.output(output.answer, output.success)
```
giordano-lucas, line 341: same here

packages/notte-agent/src/notte_agent/falco/agent.py
```diff
@@ -336,8 +352,12 @@ async def _run(self, task: str, url: str | None = None) -> AgentResponse:
                             message=failed_val_msg,
                         )
                     )
+        finally:
```
giordano-lucas, line 355: I don't like this. Let's remove that and ask the sessions to manage their sessions themselves

packages/notte-agent/src/notte_agent/falco/agent.py
```diff
@@ -150,12 +152,12 @@ async def reset(self) -> None:
         self.step_executor.reset()
         await self.session.reset()
 
-    def output(self, answer: str, success: bool) -> AgentResponse:
+    async def output(self, answer: str, success: bool) -> AgentResponse:
```
giordano-lucas, line 155: Do we need the await ?

giordano-lucas (approved):

PR #328, opened 2025-05-30
Output schema validation

leo-notte:
Make this work (and self-heal):
```python

class Product(BaseModel):
    name: str
    price: int = Field(le=5, ge=0)

agent = Agent()
valid = agent.run(
    task='CRITICAL: dont do anything, return a completion action directly with output {"name": "my name", "price": -3}. You are allowed to shift the price if it fails.',
    output_model=Product,
)
```

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying and validating custom response formats using Pydantic models or JSON schemas when running agents.
  - Agents now accept flexible keyword arguments, enabling more extensible and schema-driven requests.
  - Improved clipboard permission handling in browser automation, now conditional on browser type.

- **Bug Fixes**
  - Enhanced output validation to catch type, length, and value constraint errors in agent responses.

- **Tests**
  - Introduced comprehensive tests for schema validation, covering various constraint types and error conditions.

- **Chores**
  - Improved logging for agent execution and updated documentation for new response format handling.
  - Refactored nested asyncio enabling into a dedicated function and applied it consistently across modules.
  - Updated CLI example to use typed options and keyword arguments for improved usability.


giordano-lucas:
I think we should also probably update the prompt to send the schemato the reasoning model. Or something else that I thought about was : if the LLM selected a scrape action we say do you want to use the user provided schema to scrape the page ?

PR #389, opened 2025-06-25
Small bugfixes, batch run

leo-notte:
- batch run agents with notte.Agent.abatch
- fix llm output being none raising instantly
- reobserve when failing validation
- allow completion action answer to be given by llm as dict (converted into str by the basemodel)


packages/notte-agent/src/notte_agent/falco/agent.py
```diff
@@ -336,10 +337,13 @@ async def _run(self, request: AgentRunRequest) -> AgentResponse:
                 """
                 logger.error(failed_val_msg)
                 # add the validation result to the trajectory and continue
+
+                # need to reobserve because might loop because of missing info
+                obs = await self.session.aobserve()
```
giordano-lucas: Just merged the other PR I think we should be covered there

packages/notte-core/src/notte_core/llms/engine.py
```diff
@@ -82,7 +82,7 @@ async def structured_completion(
             tries -= 1
             try:
                 content = (
-                    await self.single_completion(messages, model, response_format=litellm_response_format)
+                    await self.single_completion(messages, model, response_format=litellm_response_format) or ""
```
giordano-lucas, line 85: are you sure about that ?

packages/notte-sdk/src/notte_sdk/endpoints/agents.py
```diff
@@ -639,6 +641,100 @@ def __init__(self, client: AgentsClient, open_viewer: Callable[[str], None]) ->
         self.client = client
         self.open_viewer = open_viewer
 
+    async def abatch(
+        self,
+        request: AgentStartRequestDict,
+        agent_create_dict: dict[str, Any],
```
giordano-lucas: can't we just use self.request instead of `agent_create_dict` ?

packages/notte-sdk/src/notte_sdk/endpoints/agents.py
```diff
@@ -639,6 +641,100 @@ def __init__(self, client: AgentsClient, open_viewer: Callable[[str], None]) ->
         self.client = client
         self.open_viewer = open_viewer
 
+    async def abatch(
+        self,
+        request: AgentStartRequestDict,
```
giordano-lucas: shouldn't we have Unpack[AgentStartRequestDict] here ?

PR #393, opened 2025-06-25
Handle svg clickable elements

leo-notte:


leo-notte:
closing, need to test further

PR #430, opened 2025-07-10
Updates to documentation and quickstart examples

shivankj11:
## Summary by CodeRabbit

* **New Features**
  * Introduced a terminal-based interactive quickstart app for running Notte AI tasks with customizable reasoning models and steps.
  * Added a command-line quickstart example script for streamlined agent execution.
  * Provided an automated Bash script for quick environment setup and example execution.
  * Added a new example script demonstrating automated web tasks with credential vault integration.
* **Documentation**
  * Simplified and clarified the README introduction and quickstart instructions, including environment setup and API key configuration.
  * Added example environment variable template for user configuration.
* **Tests**
  * Added a test to validate the new quickstart example script.


quickstart.py
```diff
@@ -0,0 +1,15 @@
+import os
+import sys
+from notte_sdk import NotteClient
```
giordano-lucas: @shivankj11 please add a test for this code before merging. We want to avoid having outdated code in the quickstart examples. But otherwise LGTM

shivankj11:
> @shivankj11 I wonder we could make the terminal UI that would allow allow users to easily allow users to select between different models, different max steps values, & parameters like the bechmark TUI of magnitude we saw yesterday

Good idea. I'll see if I can make something pretty for switching models / changing max steps

README.md
andreakiro, line 1: i think its good to have the `quickstart.sh` first (even if not common) but we should def also keep the 5 lines code snippet of how to run an agent in a Python file that you deleted. These are important for devs- as i expect they could be annoyed being forced to go through the quickstart bash script.

quickstart.sh
andreakiro, line 1: This overrides a venv if it already exists- i lost mine when running the script

PR #475, opened 2025-08-02
add togetherai and anthropic model support

omkar-notte:


packages/notte-core/src/notte_core/llms/engine.py
```diff
@@ -109,6 +109,10 @@ async def structured_completion(
             if "```json" in content:
                 # extract content from JSON code blocks
                 content = self.sc.extract(content).strip()
+            elif content.startswith('{"json":'):
```
giordano-lucas: can you put {"json": into a class variable and use content = content [len(XYZ):-1] ?

PR #478, opened 2025-08-05
Better scrape signature

giordano-lucas:
## Summary by CodeRabbit

* **New Features**
  * Enhanced scraping methods now support multiple response formats and flexible parameter handling, allowing users to retrieve markdown, structured data, or images.
  * Added a consolidated logo scraping function with improved URL normalization and fallback logic.

* **Improvements**
  * Streamlined and clarified documentation and code examples to reflect updated scraping and session execution methods.
  * Navigation steps are now explicit before scraping in tutorials and tests, improving clarity and reliability.
  * Unified action execution under a consistent `execute` method replacing previous `step` and `observe` patterns.
  * Simplified scraping calls by removing session context management in examples and tutorials.
  * Updated error handling and output for scraping actions to provide more consistent and informative results.
  * Adjusted method signatures to enforce keyword-only arguments for certain factory calls.
  * Refined scraping interface to support multiple call patterns and overloads for synchronous and asynchronous methods.
  * Improved error reporting and data handling on schema validation failures during scraping.

* **Bug Fixes**
  * Ensured markdown content is always present and non-null in scraping results.

* **Documentation**
  * Updated tutorials, SDK reference, and code snippets to match new method names and usage patterns for scraping and session actions.
  * Refined session-related documentation to reflect changes in endpoint behavior and method naming.
  * Reduced columns in session feature guides for better readability.

* **Tests**
  * Added and updated integration tests to cover new scraping workflows and response formats.
  * Standardized assertions and variable naming in scraping tests for clarity and consistency.


leo-notte (changes requested):
while you're at it i think this should go

https://github.com/nottelabs/notte/blob/6ea99e021eab68f98439f1d476b76fc2b33df004/packages/notte-browser/src/notte_browser/scraping/schema.py#L128

It doesnt make sense to me that we still return data if it doesnt match the schema, especially because get() will still return it, regardless of success

packages/notte-browser/src/notte_browser/session.py
```diff
@@ -463,22 +464,67 @@ def execute(
...
+        )
+
+    @overload
+    def scrape(self, /, **params: Unpack[ScrapeMarkdownParamsDict]) -> str: ...
+
+    @overload
+    def scrape(self, *, instructions: str, **params: Unpack[ScrapeMarkdownParamsDict]) -> StructuredData[BaseModel]: ...
+
+    @overload
+    def scrape(
```
leo-notte, line 515: thought it would be nice for debugging to still have access to the scraped data in markdown that lead to this response, but maybe this would be one more boolean param which messes with the overload

packages/notte-sdk/src/notte_sdk/types.py
```diff
@@ -1158,6 +1165,11 @@ class ScrapeParams(SdkBaseModel):
         ),
     ] = True
```
leo-notte, line 1166: i wonder if this shouldn't be set to false by default, especially because of some iframes being handled weirdly when set to true

leo-notte:
> while you're at it i think this should go
> 
> https://github.com/nottelabs/notte/blob/6ea99e021eab68f98439f1d476b76fc2b33df004/packages/notte-browser/src/notte_browser/scraping/schema.py#L128
> 
> It doesnt make sense to me that we still return data if it doesnt match the schema, especially because get() will still return it, regardless of success

example output:

 
image

script:
```
from pydantic import BaseModel

import notte


class PricingPlan(BaseModel):
    name: str
    price_per_month: int | None
    features: list[str]
    api_credits: int
    num_vaults: int


class PricingPlans(BaseModel):
    plans: list[PricingPlan]


with notte.Session() as session:
    session.execute(type="goto", value="https://www.notte.cc")
    data = session.scrape(
        response_format=PricingPlans,
        instructions="Extract the pricing plans from the page. try to fetch as much data as possible, especially for num_vaults and api credits, those are necessary",
    )

    import logging

    logging.warning(f"{data=}")
```

packages/notte-sdk/src/notte_sdk/types.py
giordano-lucas: I don't know I wouldn't really update that without benchmarking this first since this has been setup for a while now

packages/notte-browser/src/notte_browser/session.py
giordano-lucas: I agree but at the same time, it's very easy to just do another scrape without any parameter to debug. Or maybe we could just user `logger.debug(space.markdown)` and that's it

packages/notte-sdk/src/notte_sdk/types.py
leo-notte: I guess, i suppose it's also dependent on if you also have the schema scraping pipe running or not

packages/notte-browser/src/notte_browser/session.py
leo-notte: true, i guess since scraping is deterministic i'd be fine with that

PR #492, opened 2025-08-15
Add save full video option to session

shivankj11:
notte.Session(record_video_dir='...')


packages/notte-sdk/src/notte_sdk/types.py
```diff
@@ -498,6 +500,13 @@ class SessionStartRequest(SdkBaseModel):
         False
     )
 
+    # Video recording options
+    record_video_dir: Annotated[str | None, Field(description="Directory to save recorded videos to.")] = None
```
giordano-lucas, line 504: I wouldn't put that in there since this will also allow users to use these parameters in the SDK but this should not be supported there. You can put that in the contsructor of the notte.Session directly

giordano-lucas (changes requested):
Cf previous comment

PR #511, opened 2025-08-21
[WIP] raw file download handling

omkar-notte:


packages/notte-browser/src/notte_browser/controller.py
```diff
@@ -1,3 +1,4 @@
+import requests
```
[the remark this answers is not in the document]
giordano-lucas: @omkar-notte you might want to do that

packages/notte-core/src/notte_core/utils/raw_file.py
```diff
@@ -0,0 +1,54 @@
...
+
+from notte_core.browser.dom_tree import ComputedDomAttributes, DomAttributes, DomNode, NodeSelectors
+from notte_core.browser.node_type import NodeRole, NodeType
+
+
+def get_file_ext(headers: dict[str, Any]) -> str:
+    return mimetypes.guess_extension(headers["content-type"]) or ""
+
+
+def get_filename(headers: dict[str, Any], url: str) -> str:
```
giordano-lucas, line 18: can you add some tests also for this function ? With different headers type so that every if/else case is covered

omkar-notte: added `test_raw_file.py`

PR #539, opened 2025-09-02
enable steps for session, agent from id init, mini fixes

leo-notte:
## Summary by CodeRabbit

- New Features
  - Remote agents can be created from a session or referenced by agent ID; batched runs honor both modes and expose agent/session identifiers with safeguards for unsupported actions.

- Bug Fixes
  - Improved fill action messages to omit awkward spacing when labels are missing, yielding cleaner logs.

- Tests
  - Added integration tests for new and legacy session/agent status formats and step sequences.

### DIFF
--- a/packages/notte-sdk/src/notte_sdk/endpoints/agents.py
+++ b/packages/notte-sdk/src/notte_sdk/endpoints/agents.py
@@ -848,7 +848,7 @@ class RemoteAgent:
             AgentResponse: The initial response from starting the agent.
         """
         if self.existing_agent:
-            raise ValueError("You cannot run an existing agent")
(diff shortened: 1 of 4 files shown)


packages/notte-sdk/src/notte_sdk/endpoints/agents.py
```diff
@@ -810,6 +847,9 @@ def start(self, **data: Unpack[AgentRunRequestDict]) -> AgentResponse:
         Returns:
             AgentResponse: The initial response from starting the agent.
         """
+        if self.existing_agent:
+            raise ValueError("You cannot run an existing agent")
```
giordano-lucas: better error message ?

PR #586, opened 2025-09-30
Small changes to docs

leo-notte:


docs/src/concepts/workflows.mdx
```diff
@@ -17,8 +17,6 @@ Notte workflows are simple python scripts that can be executed both locally and
 ## Workflow Management
 ### Python SDK
 
-The following snippet shows how to manage your workflows using the Notte Python SDK.
```
giordano-lucas, line 20: intended change ?

packages/notte-sdk/src/notte_sdk/actions.py
```diff
@@ -0,0 +1,112 @@
+from notte_core.actions import (
```
giordano-lucas, line 1: can't we do that automatically using a for loop ? I feel like we are going to forget to update this file everytime we add a new action

docs/src/sdk-reference/misc/checkaction.mdx
```diff
@@ -7,13 +7,10 @@ Use `True` to check, `False` to uncheck.
...
-from notte_core.actions import CheckAction
-# Dict syntax
-session.execute({"type": "check", "id": "terms-checkbox", "value": True})
-session.execute({"type": "check", "id": "newsletter-checkbox", "value": False})
-# Pydantic syntax
-session.execute(CheckAction(id="terms-checkbox", value=True))
-session.execute(CheckAction(id="newsletter-checkbox", value=False))
+from notte_sdk import actions
+
+session.execute(actions.Check(id="terms-checkbox", value=True))
```
giordano-lucas, line 12: Can we set `selector=` instead of `id=` ?

packages/notte-sdk/src/notte_sdk/actions.py
leo-notte: wanted to do it that way at first, but it wasn't super clear, so i just added a test to make sure that all actions are imported

PR #591, opened 2025-10-06
add playwright page in notte-sdk

giordano-lucas:
## Summary by CodeRabbit

- **New Features**
  - Optional Playwright support: access a Chromium page via CDP with lazy initialization for sync and async sessions; browser resources are set up and cleaned up on session enter/exit (no change unless Playwright is installed).

- **Chores**
  - Added an optional Playwright dependency group (playwright>=1.50.0).
  - Relaxed import checks to permit Playwright-related code in designated files.

- **Bug Fixes**
  - Tightened scraping test assertions to be robust against surrounding whitespace and simplified expected content.

### DIFF
--- a/packages/notte-sdk/src/notte_sdk/endpoints/sessions.py
+++ b/packages/notte-sdk/src/notte_sdk/endpoints/sessions.py
@@ -998,7 +998,8 @@ class RemoteSession(SyncResource):
         try:
             # Initialize playwright context
             if self._playwright_context is None:
-                assert _sync_playwright is not None
+                if _sync_playwright is None:
+                    raise RuntimeError("Playwright is not initialized")
                 self._playwright_context = _sync_playwright().start()
 
             # Connect to browser via CDP
@@ -1052,7 +1053,8 @@ class RemoteSession(SyncResource):
         try:
             # Initialize async playwright context
             if self._async_playwright_context is None:
-                assert _async_playwright is not None
+                if _async_playwright is None:
+                    raise RuntimeError("Playwright is not initialized")
                 self._async_playwright_context = await _async_playwright().start()
 
             # Connect to browser via CDP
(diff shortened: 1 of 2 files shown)


packages/notte-sdk/src/notte_sdk/endpoints/sessions.py
```diff
@@ -883,6 +951,101 @@ def cdp_url(self) -> str:
...
+            assert _sync_playwright is not None
+            self._playwright_context = _sync_playwright().start()
+
+        # Connect to browser via CDP
+        if self._playwright_browser is None:
+            cdp_url = self.cdp_url()
+            self._playwright_browser = self._playwright_context.chromium.connect_over_cdp(cdp_url)
+
+        # Get the first page from the first context
+        self._playwright_page = self._playwright_browser.contexts[0].pages[0]  # pyright: ignore
```
leo-notte: might just wanna have an error message around here just in case something fails and we have no context or page

packages/notte-sdk/src/notte_sdk/endpoints/sessions.py
```diff
@@ -883,6 +951,101 @@ def cdp_url(self) -> str:
...
+        """
+        if not _async_playwright_available:
+            raise ImportError("Playwright not installed. Use `pip install notte-sdk[playwright]` to install it.")
+
+        # Return cached page if already connected
+        if self._async_playwright_page is not None:
+            return self._async_playwright_page
+
+        # Initialize async playwright context
+        if self._async_playwright_context is None:
```
leo-notte: i feel like even if we use the async with for the session, we'll still be able to get the sync page, but they won't get cleaned up during exit (and vice versa?)

maybe it works out fine because of the playwright message when in async context, but maybe worth checking

PR #736, opened 2026-03-11
fix: agent fallback validator

leo-notte:
## Summary by CodeRabbit

* **New Features**
  * Action execution and scrape results now include URL context so outputs show which page an action or scrape ran on.
  * Enhanced system guidance clarifying behavior across multi-page navigation and element ID resets.

* **Tests**
  * Added comprehensive tests for URL tracking and multi-page navigation, plus an integration-marked placeholder test.

* **Chores**
  * Test config updated to support an "integration" marker.

<h3>Greptile Summary</h3>

This PR fixes a bug where the `CompletionValidator` would incorrectly reject a valid agent completion after page navigation. When an agent clicked a link that navigated to a new page, the validator saw the new page's elements but the action history referenced elements from the previous page — leading to false negatives. The fix threads the page URL at time of execution through the data model (`ExecutionResult.url`), surfaces it in the agent's perception layer (`FalcoPerception.perceive_action_result`), and adds a clear multi-page context explanation to the validator system prompt.

**Key changes:**
- `observation.py` — adds an optional `url` field to `ExecutionResult` (backward compatible, defaults to `None`).
- `session.py` — captures `self.page.url` before both action execution and scrape execution, and passes it to every newly created `ExecutionResult`.
- `perception.py` — appends `(on <url>)` to the formatted action-result string when a URL is available, giving the LLM provenance for each action in the history.
- `validator.py` — extends the system prompt with an `IMPORTANT` block explaining that element IDs reset per page and that action history URLs reflect where actions were actually performed.
- New tests cover `ExecutionResult` URL population, `perceive_action_result` output format, system prompt content, and end-to-end validation after page navigation.

**Concerns:**
- `test_validator_accepts_completion_after_page_navigation` in `tests/agent/test_validator.py` is a full end-to-end test (real browser, LLM, and external network calls to `example.com`/`iana.org`) with no `pytest.mark.integration` guard or skip condition. It will run with the unit-test suite in any environment and fail in offline/resource-constrained CI.
- Inside `test_validator_receives_url_in_action_history`, `asyncio.run()` is called inside a synchronous test to append to a trajectory. This can conflict with an already-running event loop in some pytest configurations.

<h3>Confidence Score: 4/5</h3>

- Safe to merge — the core fix is correct, backward compatible, and well-tested; only the test hygiene concerns need attention.
- The data-model change (`url` field with `None` default) is fully backward compatible. URL capture happens at exactly the right moment (before execution) in both the action and scrape code paths. The perception and validator prompt changes are additive. The only risks are a fragile end-to-end test placed outside the integration folder (no `pytest.mark` guard) and potential `asyncio.run` conflicts in the test suite — neither affects production behaviour.
- tests/agent/test_validator.py — the new end-to-end test `test_validator_accepts_completion_after_page_navigation` should be guarded with `pytest.mark.integration` or moved to `tests/integration/`.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| packages/notte-core/src/notte_core/browser/observation.py | Adds an optional `url: str | None = None` field to `ExecutionResult` to record the page URL at the time of action execution. Backward compatible with a `None` default. |
| packages/notte-browser/src/notte_browser/session.py | Captures `self.page.url` before both action execution and scrape execution, then passes it to all created `ExecutionResult` objects via the new `url` field. Correctly covers both the action path and the scrape path. |
| packages/notte-agent/src/notte_agent/falco/perception.py | Adds `(on <url>)` to the success and failure action-result strings when a URL is present, providing the LLM with provenance context for each action in the history. |
| packages/notte-agent/src/notte_agent/common/validator.py | Adds an IMPORTANT block to the validator system prompt explaining that actions were performed on previous pages, that the current observation is the result of navigation, and that element IDs reset per page. |
| tests/agent/test_validator.py | Adds several unit tests for the new URL field and one end-to-end test (`test_validator_accepts_completion_after_page_navigation`) that requires a real browser, LLM, and external network access; this test is placed in the non-integration test folder without a marker and without a skip guard, making it potentially fragile in offline/CI environments. |
| tests/integration/sdk/test_agent_fallback.py | Adds a stub test that immediately calls `pytest.skip`, redirecting testers to the local test in `test_validator.py`. Clean placeholder with a useful message. |

</details>

<sub>Last reviewed commit: 3e35d68</sub>

> Greptile also left **2 inline comments** on this PR.

### DIFF
--- a/packages/notte-agent/src/notte_agent/common/validator.py
+++ b/packages/notte-agent/src/notte_agent/common/validator.py
@@ -17,6 +17,11 @@ If the task is unclear defined, you can let it pass.
 But if something is missing or the image does not show what was requested dont let it pass.
 Try to understand the page and help the model with suggestions like scroll, do x, ... to get the solution right.
 
+IMPORTANT: The agent navigates through multiple pages. Actions in the history were executed
+on previous pages (check the URL in each action). The current page observation shows the
+RESULT of navigation, not where actions were performed. Element IDs (L1, L2, etc.) reset
+on each page - do not assume IDs in action history match current page elements.
+
 Task to validate: {{task}}.
 
 Return a JSON object with 2 keys: `is_valid` and `reason`:
--- a/packages/notte-agent/src/notte_agent/falco/perception.py
+++ b/packages/notte-agent/src/notte_agent/falco/perception.py
@@ -103,10 +103,13 @@ Data scraped from current page view:
         include_data: bool = True,
     ) -> str:
         id_str = f" with id={result.action.id}" if include_ids and isinstance(result.action, InteractionAction) else ""
+        url_str = f" (on {result.url})" if result.url else ""
         if not result.success:
             err_msg = trim_message(result.message)
-            return f"❌ action '{result.action.name()}'{id_str} failed with error: {err_msg}"
-        success_msg = f"✅ action '{result.action.name()}'{id_str} succeeded: '{result.action.execution_message()}'"
+            return f"❌ action '{result.action.name()}'{id_str} failed{url_str}: {err_msg}"
+        success_msg = (
+            f"✅ action '{result.action.name()}'{id_str} succeeded{url_str}: '{result.action.execution_message()}'"
+        )
         if include_data:
             return f"{success_msg}{self.perceive_data(result.data, only_structured=True)}"
         return success_msg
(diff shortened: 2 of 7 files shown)


eeshsaxena (commented):
Reads well. Capturing `action_url = self.page.url` before the action runs is the right call, since the goal is to record where each action was performed rather than where navigation landed afterward. Surfacing that as `(on {url})` in `perceive_action_result`, together with the new prompt note that element IDs reset per page, gives the validator the context to stop conflating history actions with the current observation. Low-risk and self-consistent.

PR #786, opened 2026-04-24
fix(agent): stop hangs on CDP detach and stalled LLM calls

leo-notte:
## Summary

Two independent fixes that together eliminate the "node down: Not properly terminated" worker crashes seen in recent CI runs (e.g. run [24903852728](https://github.com/nottelabs/notte/actions/runs/24903852728)).

Both failure modes looked identical from CI: a test hangs for exactly 300s, the pytest-timeout thread method calls `os._exit(1)` to kill the worker, xdist reports "Not properly terminated", the test is rerun, cycle repeats. Under the hood they were two different bugs in the same code path.

### 1. CDP session attach/detach race on navigation

`_cdp_screenshot` was creating a new CDP session on every screenshot via `context.new_cdp_session(page)` and detaching it in a `finally` block. CDP sessions on a page survive navigations within that page, so the per-call detach was both unnecessary and actively harmful: it raced with Chromium's target lifecycle after a navigation, leaving the next operation parked on a dead socket.

Fix: cache one CDP session per page on the window, reuse across calls, drop on error, detach once in `close()`.

### 2. Unbounded LLM HTTP call

`litellm.acompletion(...)` was called with no `timeout` kwarg. When the upstream HTTP stream stalled (no data, no FIN/RST), httpx waited forever and the event loop parked.

Fix: `timeout=60`. Stalls now raise `litellm.Timeout` cleanly.

## Evidence

Reproduced the hang locally, isolated each source, and confirmed both fixes hold in parallel runs against localhost and against us-dev.notte.cc staging.

| metric | before | after |
|---|---|---|
| CDP detach warnings | present | 0 |
| Worker deaths in batch | yes | 0 |
| 300s zombie hangs | yes | 0 |
| LLM stall behaviour | hang forever | raise `Timeout` at 60s |

## Test plan

- [ ] CI run is green (or at least, the previously-hanging tests no longer wedge workers)
- [ ] Local run of browser-heavy agent tests does not exhibit `+++ Timeout +++` traces
- [ ] `_cdp_screenshot` still produces valid screenshots across page navigations (covered by existing screenshot tests)
- [ ] `litellm.Timeout` surfaces as an agent-level error rather than hanging

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced overhead and improved reliability of browser window captures by reusing debugging sessions, improving session lifecycle, and enhancing error recovery during captures
  * Added a 60-second timeout for LLM completion requests to avoid indefinite waits

* **Tests**
  * Preserved two additional important persona IDs in persona-management tests
  * Marked flaky integration tests to allow automatic reruns on transient failures

---

> [!NOTE]
> Fixes two independent causes of agent run hangs: CDP session caching per page (replacing per-call create/detach that raced with Chromium's target lifecycle), and a 60s timeout on `litellm.acompletion` to prevent stalled HTTP streams from parking the event loop. Also adds two deterministic-seeded persona UUIDs to the cleanup allow-list and marks `test_new_steps` and `test_agent_fallback` as flaky.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit d802b2fe9fb5a45ccd24bccee02cb478c131ea7b.</sup>

### DIFF
--- a/packages/notte-browser/src/notte_browser/window.py
+++ b/packages/notte-browser/src/notte_browser/window.py
@@ -29,7 +29,7 @@ from notte_sdk.types import (
     Cookie,
     SessionStartRequest,
 )
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, PrivateAttr
 from typing_extensions import override
 
 from notte_browser.dom.parsing import dom_tree_parsers
@@ -170,6 +170,12 @@ class BrowserWindow(BaseModel):
 
     model_config: ClassVar[ConfigDict] = ConfigDict(arbitrary_types_allowed=True)
 
+    # Cache of CDP sessions keyed by id(page). CDP sessions survive navigations on
+    # the same Page target, so creating/detaching per call raced with Chromium's
+    # target lifecycle after navigation and could park subsequent operations on a
+    # dead socket.
+    _cdp_sessions: dict[int, CDPSession] = PrivateAttr(default_factory=dict)
+
     @override
     def model_post_init(self, __context: Any) -> None:
         self.resource.page.set_default_timeout(config.timeout_default_ms)
@@ -206,6 +212,7 @@ class BrowserWindow(BaseModel):
     async def close(self) -> None:
         if self.on_close is not None:
             await self.on_close()
+        await self._close_all_cdp_sessions()
         for page in self.resource.page.context.pages:
             if not page.is_closed():
                 await page.close()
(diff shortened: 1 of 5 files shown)


leo-notte:
@greptileai review

leo-notte:
@greptileai review

leo-notte:
@greptileai please review the latest commits and post an updated review comment with confidence score and any new issues found.

leo-notte:
@greptileai  review the latest commits and post a NEW comment with confidence score and any new issues found.

leo-notte:
@greptileai please review the latest commits and update your summary comment with the new confidence score.

PR #797, opened 2026-05-04
fix(credentials): raise when sentinel placeholder hits an invalid element

leo-notte:
## Summary

When a known credential sentinel (e.g. `PASSWORD = "mycoolpassword"` from `notte_core.credentials`) was supplied for a `FillAction` but the targeted element failed the credential field's `validate_element` check, `replace_credentials` silently logged at trace level and returned the action unmodified — causing the **literal placeholder string** to be typed into the form.

This was easy to hit by targeting a `<label>` rather than its associated `<input type="password">` (e.g. via observe IDs that point at the label):

- `UserNameField.validate_element` returns `True` unconditionally → username substitution proceeded.
- `PasswordField.validate_element` requires `attrs.type == "password"` → password substitution silently no-op'd while the fill itself reported success.

End result: `mycoolpassword` got typed into the password field as plaintext and the user only noticed because they manually re-read `.value` via `eval-js`.

### Changes

- **`notte_core.errors.processing`**: new `CredentialFieldValidationError(NotteBaseError)` with dev/agent/user messages naming the placeholder, credential key, and offending element attrs.
- **`notte_core.credentials.base.replace_credentials`**: replaces the silent `logger.trace` with a raise. The branch only fires when the value already matched a known sentinel (lines 658–661 still raise `InvalidPlaceholderError` for unknown values), so raising here is the correct signal: the caller clearly intended a substitution and the literal sentinel must not leak into typing.
- **`notte_browser.session._action_with_vault`**: adds a dedicated `except CredentialFieldValidationError: raise` ahead of the broad `except ValueError`, otherwise the new error would be swallowed (since `NotteBaseError` extends `ValueError`).
- **`tests/test_session_vault_helpers.py`**: 4 new tests covering the new raise, the happy substitute path, the asymmetric username path, and propagation through the session-level catch.

### Out of scope

- The asymmetric design of `validate_element` (username unconditionally accepts, password requires `type=="password"`) is preserved — pinned by the new `test_username_placeholder_unaffected_by_validation_change`. Tightening username validation would be a separate behavioural change.
- Resolving label → associated input before reading `LocatorAttributes` would also fix the reported case but is a larger refactor; the raise is the safer first step because it surfaces the failure instead of corrupting the field.

## Test plan

- [x] `pytest tests/test_session_vault_helpers.py` — 8 passed (4 existing + 4 new)
- [ ] Manual: rerun the original repro (`notte page fill M2 mycoolpassword` on `the-internet.herokuapp.com/login` with a vault entry) and confirm the call now raises instead of silently typing `mycoolpassword` into the password field

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential field validation to detect and report targeting errors immediately. When credential placeholders target incompatible form fields (e.g., password credentials on non-password fields), the system now raises explicit validation errors instead of silently continuing and failing later, providing clearer error messages for debugging misconfigured credentials.

* **Tests**
  * Added comprehensive validation test coverage for credential field targeting scenarios, including password and username field validation and error propagation.

---

> [!NOTE]
> Replaces a silent `logger.trace` no-op in `replace_credentials` with a new `CredentialFieldValidationError` when a known credential sentinel targets an element that fails `validate_element`. Adds a dedicated `except CredentialFieldValidationError: raise` in `_action_with_vault` to prevent the broad `except ValueError` from swallowing it. Four new tests cover the error path, happy path, asymmetric username behaviour, and session-level propagation.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 52fbcdbf39c484d8b321cba0ee84df1c5d529a75.</sup>

### DIFF
--- a/packages/notte-browser/src/notte_browser/session.py
+++ b/packages/notte-browser/src/notte_browser/session.py
@@ -65,6 +65,7 @@ from notte_core.credentials.base import BaseVault, LocatorAttributes
 from notte_core.data.space import DataSpace, ImageData, StructuredData, TBaseModel
 from notte_core.errors.actions import InvalidActionError
 from notte_core.errors.base import NotteBaseError
+from notte_core.errors.processing import CredentialFieldValidationError
 from notte_core.errors.provider import RateLimitError
 from notte_core.profiling import profiler
 from notte_core.space import ActionSpace
@@ -545,6 +546,10 @@ class NotteSession(AsyncResource, SyncResource):
                     outerHTML=await locator.evaluate("el => el.outerHTML"),
                 )
             return await self.vault.replace_credentials(action, attrs, snapshot)
+        except CredentialFieldValidationError:
+            # Sentinel placeholder used on an invalid target element — surface loudly instead of
+            # silently typing the literal placeholder string into the field.
+            raise
         except ValueError as e:
             # Credential field not found in vault (e.g., vault has email but action needs username)
             # Return original action - it will fail at execution with a clearer error
--- a/packages/notte-core/src/notte_core/credentials/base.py
+++ b/packages/notte-core/src/notte_core/credentials/base.py
@@ -26,7 +26,7 @@ from notte_core.common.logging import logger
 from notte_core.common.types import TResponseFormat
 from notte_core.credentials.types import ValueWithPlaceholder, get_str_value
 from notte_core.errors.actions import NoCredentialsFoundError
-from notte_core.errors.processing import InvalidPlaceholderError
+from notte_core.errors.processing import CredentialFieldValidationError, InvalidPlaceholderError
 from notte_core.profiling import profiler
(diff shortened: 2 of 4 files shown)


leo-notte:
@greptileai review

eeshsaxena (commented):
Good fix. Since `NotteBaseError` subclasses `ValueError`, the new `CredentialFieldValidationError` would otherwise be caught by the `except ValueError` in `_action_with_vault` and downgraded to the "field not found, return original action" path, which is exactly the silent no-op that let the literal sentinel placeholder get typed into the field. Putting `except CredentialFieldValidationError: raise` ahead of the `except ValueError` is what makes the loud failure actually propagate.

Moving from `logger.trace(...)` to raising only happens in the `else` branch where `validate_element` already failed, so the success path is unchanged; this just turns a silent miss into a clear, actionable error. Reads correct.

PR #802, opened 2026-05-07
fix: stop silently swallowing exceptions in forward_incremental

dashitongzhi:
## Summary

The retry loop in `forward_incremental` was catching all exceptions with a bare `except Exception: pass`, silently discarding all error information. This made it impossible to diagnose why incremental action listing was failing.

## Changes

In `RetryPipeWrapper.forward_incremental()`:

- **Collect error messages** in an `errors` list (matching the pattern in `forward()`)
- **Log each failure** when verbose mode is on, showing the first 200 chars of the error message
- **Trace failures** via `self.tracer.trace()` so they appear in telemetry/debug output
- **Improved log message** when max retries are exhausted: now includes retry count and all error messages

## Before
```python
for _ in range(self.max_tries):
    try:
        return await self.pipe.forward_incremental(snapshot, previous_action_list)
    except Exception:
        pass  # 🔇 all errors silently swallowed
```

## After
```python
errors: list[str] = []
last_error: Exception | None = None
for _ in range(self.max_tries):
    try:
        return await self.pipe.forward_incremental(snapshot, previous_action_list)
    except Exception as e:
        last_error = e
        errors.append(str(e))
        if self.verbose:
            logger.debug(f"forward_incremental: failed ... error msg: {str(e)[:200]}...")
```

This matches the error handling pattern already used in the sibling `forward()` method.

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Enhanced retry behavior for incremental processing by capturing per-attempt error messages and increasing visibility into failures with more informative verbose debug logs.
  * When all retries are exhausted, the system emits a failure event containing retry details and falls back to the prior action set, preserving prior behavior while making issues easier to diagnose.

---

> [!NOTE]
> Fixes silent exception swallowing in `forward_incremental` by collecting errors, logging them in verbose mode, and reporting failures to the tracer — matching the pattern already used in `forward()`. A follow-up commit prefixes `last_error` with `_` to resolve the ruff F841 warning and reformats the file.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit c238a3a0cf8a1d3e64820ffff0aaab0ba4a7e985.</sup>


leo-notte:
Closing as superseded by #868, which has now been merged. It includes the forward_incremental observability fix while preserving the fallback behavior and refining retry handling. Thank you @dashitongzhi for identifying the issue and providing the original implementation.

PR #820, opened 2026-05-12
Fix headless aspect ratio viewport handling

leo-notte:
## Summary
- preserve `aspect_ratio` through `BrowserWindowOptions` so local headless sessions can derive an aspect-matching viewport instead of falling back to the default headless size
- clear implicit SDK viewport defaults when `aspect_ratio` is provided, while still rejecting explicit viewport dimensions combined with `aspect_ratio`
- avoid synthesizing viewport dimensions for headless CDP sessions so remote browser geometry/fingerprint state is preserved

## Why
Remote browser-pod sessions already apply coherent screen/avail/window geometry in the browser service before the worker connects over CDP. Synthesizing a viewport in the worker causes Playwright/CDP viewport emulation to collapse `screen.*` to the viewport size, creating inconsistent fingerprint geometry.

## Tests
- `uv run pytest tests/browser/test_window_options.py`

## Summary by CodeRabbit

* **New Features**
  * Added aspect-ratio support for headless browser windows, calculating viewport height from a default headless width (e.g., 16:9, 5:4).
  * When a remote/CDP endpoint is used in headless mode, the viewport is left unset.

* **Bug Fixes**
  * Explicit viewport values continue to override aspect-ratio defaults.
  * Validation now ensures aspect-ratio presets are applied exclusively (clears explicit viewport fields).

* **Tests**
  * Added unit and integration tests covering aspect-ratio behavior and precedence rules.

---

> [!NOTE]
> Adds `aspect_ratio` support to `BrowserWindowOptions` so headless sessions derive viewport dimensions from an aspect ratio string instead of falling back to the randomised default. CDP sessions short-circuit before viewport synthesis to preserve remote browser geometry. The SDK validator is updated to clear implicit viewport defaults when `aspect_ratio` is provided, and new unit + integration tests cover both paths.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit dca59bcd3417d6887de560494c398b5e90570a62.</sup>

### DIFF
--- a/packages/notte-browser/src/notte_browser/window.py
+++ b/packages/notte-browser/src/notte_browser/window.py
@@ -79,6 +79,12 @@ class BrowserWindowOptions(BaseModel):
             raise ValueError("Both viewport_width and viewport_height must be set together or both must be None")
 
         if self.headless and self.viewport_width is None and self.viewport_height is None:
+            if self.cdp_url is not None:
+                logger.info(
+                    "🪟 Headless CDP session detected. Leaving viewport unset so the remote browser geometry is preserved."
+                )
+                return
+
             if self.aspect_ratio is not None:
                 self.viewport_width, self.viewport_height = self._fit_aspect_ratio_in_default_viewport(
                     self.aspect_ratio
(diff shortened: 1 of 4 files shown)


leo-notte:
@greptileai review

PR #846, opened 2026-05-26
docs: add Fireworks AI integration page

sam-notte:
## Summary

- Adds `docs/src/integrations/fireworks.mdx` documenting how to run Notte agents on Fireworks-served models (Kimi K2.5, GLM-5, MiniMax M2.5)
- Registers the page in `docs/src/docs.json` under the Integrations group (placed after `openai-cua`)
- Modelled on the existing `openai-cua.mdx` format: frontmatter + `AgentMdNotice` partial + `<Steps>` setup + validated-models table + reproducibility block + resources

## Why now

Companion docs page for the Fireworks ecosystem listing ask. Their [agent-frameworks page](https://docs.fireworks.ai/ecosystem/integrations/agent-frameworks) links out to each framework's own Fireworks integration docs (LangChain, LlamaIndex, and Pydantic all follow this pattern). Notte needs its own equivalent page live before pinging them to add Notte to the listing, otherwise the card has nothing to link to.

## Verified for merge

- All three benchmark model IDs are now present in the validated-models table (`kimi-k2p5`, `glm-5`, `minimax-m2p5`), verified against the [Fireworks model catalogue](https://fireworks.ai/models).
- The joint benchmark URL now points to the live post at `https://fireworks.ai/blog/agent-execution-tax`.
- The Notte explainer link at `https://notte.cc/blog/agent-execution-tax` is live (published 2026-05-19).

## Test plan

- [ ] Mintlify preview renders the page under Integrations
- [ ] External fireworks.ai links resolve
- [ ] Internal link to /blog/agent-execution-tax resolves

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Fireworks AI integration guide covering setup prerequisites, installation steps, configuration, supported models (Kimi K2.5, GLM-5, MiniMax M2.5), and usage examples for Notte browser agents.
  * Included code snippets and resources for getting started with Fireworks integration.

PR #851, opened 2026-06-10
feat(core): add Fireworks AI as an LLM provider

sam-notte:
## Summary

Adds Fireworks AI as a supported LLM provider in `notte-core`, with three `LlmModel` aliases for the Fireworks-hosted models validated in the joint Notte × Fireworks benchmark (Kimi K2.5, GLM-5, MiniMax M2.5).

## Why

Companion to the Fireworks docs PR at #846, which documents the integration and uses `reasoning_model` strings of the form `fireworks_ai/accounts/fireworks/models/<slug>`. Without this change, the cloud rejects those strings at `/agents/start` with a 422 because the `LlmProvider` enum has no `fireworks_ai` entry, so `LlmModel.get_provider()` raises and the validator returns "not supported."

Empirically confirmed by running `agent.run()` against the cloud with `reasoning_model="fireworks_ai/accounts/fireworks/models/kimi-k2p5"`:

> 422: Model 'fireworks_ai/accounts/fireworks/models/kimi-k2p5' is not supported. Available models are: {openai/gpt-4o, ..., moonshot/kimi-k2.5, minimax/minimax-m2.5, ...}

Cloud validation reads its allowlist via `LlmModel.valid()` in this same file, so this single SDK change unlocks the docs example once the package is deployed and `FIREWORKS_API_KEY` is set in the cloud's environment.

## Diff

`packages/notte-core/src/notte_core/common/config.py`:

- `LlmProvider.fireworks_ai = "fireworks_ai"`
- `apikey_name` match arm returning `"FIREWORKS_API_KEY"`
- Three `LlmModel` aliases:
  - `fireworks_kimi_k2p5 = "fireworks_ai/accounts/fireworks/models/kimi-k2p5"`
  - `fireworks_glm_5 = "fireworks_ai/accounts/fireworks/models/glm-5"`
  - `fireworks_minimax_m2p5 = "fireworks_ai/accounts/fireworks/models/minimax-m2p5"`
- Adds `fireworks_ai` to `unsupported_providers` in `use_strict_response_format`, matching the benchmark configuration (which used non-strict `json_object`)

## Cloud requirement

After merge + redeploy, the cloud needs `FIREWORKS_API_KEY` set in its production environment for these models to appear in `LlmModel.valid()`. Without that env var, `has_apikey_in_env()` filters them out.

## Naming

The model aliases (`fireworks_kimi_k2p5`, `fireworks_glm_5`, `fireworks_minimax_m2p5`) follow the convention of prefixing with the provider to avoid collision with the existing `kimi2_5` (Moonshot direct) and `minimax` (MiniMax direct) entries. Happy to rename to any preferred convention.

## Test plan

- [ ] CI passes (syntax, type check, unit tests)
- [ ] After merge + cloud redeploy with `FIREWORKS_API_KEY` set, `agent.run(reasoning_model="fireworks_ai/accounts/fireworks/models/kimi-k2p5")` succeeds against the cloud
- [ ] Docs PR #846 example renders and runs end to end

## Summary by CodeRabbit

* **New Features**
  * Added Fireworks AI as an LLM provider (configured via `FIREWORKS_API_KEY`)
  * Added Fireworks-backed models: Kimi K2.5, GLM 5, and Minimax M2.5

* **Behavior Changes**
  * Fireworks-backed models do not enforce strict response formatting
  * Kimi K2.5 now uses a temperature override of `1.0`

* **Tests**
  * Added coverage for provider registration, model resolution, response-format behavior, and temperature override

### DIFF
--- a/packages/notte-core/src/notte_core/common/config.py
+++ b/packages/notte-core/src/notte_core/common/config.py
@@ -240,6 +240,7 @@ class LlmModel(StrEnum):
         # Model-specific temperature overrides
         temperature_overrides: dict[str, float] = {
             "kimi-k2.5": 1.0,
+            "kimi-k2p5": 1.0,
         }
         for model_pattern, temp in temperature_overrides.items():
             if model_pattern in val_str:
--- a/packages/notte-core/src/notte_core/common/config.py
+++ b/packages/notte-core/src/notte_core/common/config.py
@@ -71,6 +71,7 @@ class LlmProvider(StrEnum):
     zai = "zai"
     qwen = "qwen"
     minimax = "minimax"
+    fireworks_ai = "fireworks_ai"
 
     @property
     def context_length(self) -> int:
@@ -121,6 +122,8 @@ class LlmProvider(StrEnum):
                 return "QWEN_API_KEY"
             case LlmProvider.minimax:
                 return "MINIMAX_API_KEY"
+            case LlmProvider.fireworks_ai:
+                return "FIREWORKS_API_KEY"
             case _:  # pyright: ignore[reportUnnecessaryComparison]
                 raise ValueError(f"No API key name found for provider: {self}")  # pyright: ignore[reportUnreachable]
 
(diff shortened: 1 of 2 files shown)


sam-notte:
Update after review: ran all three slugs against Fireworks direct with a real key. `glm-5` and `minimax-m2p5` both 404 ("not deployed") — rotated off serverless since the benchmark; only `kimi-k2p5` still serves.

`108a9dc`: bumped the two dead slugs → `glm-5p2` / `minimax-m3` (both verified 200), renamed aliases to `fireworks_kimi/_glm/_minimax` (version dropped from the key so future rotations only change the value), test updated. Note the live versions aren't the exact ones benchmarked (those are gone) — same dead slugs exist in docs PR #846.

sam-notte:
Re: renaming `fireworks_ai` → `fireworks` — has to stay `fireworks_ai`: that's litellm's provider key and the model-id prefix (`fireworks_ai/...`), there's no bare `fireworks` provider. And this PR is the **direct** Fireworks path, not OpenRouter — the "fireworks" on openrouter.ai is OpenRouter's own namespace (you'd call `openrouter/z-ai/glm-5` there, not `fireworks_ai/...`). So the slugs here are correct for direct, which is what was tested.

giordano-lucas:
@greptile new review

Issue #854, opened 2026-06-24
Fireworks: allowlist by provider, not per-model enum


sam-notte:
**Problem.** Fireworks models are pinned as individual `LlmModel` enum entries, and the cloud uses `LlmModel.valid()` as its allowlist — so every model rotation needs an SDK code change + release. This just bit us: `glm-5` / `minimax-m2p5` (shipped in #851, documented in #846) 404'd within months, replaced by `glm-5p2` / `minimax-m3`.

**Proposal.**
1. Allow any `fireworks_ai/accounts/fireworks/models/*` through validation (provider-level), so new models work without an SDK bump.
2. Single source of truth for example slugs so docs don't drift from the SDK independently (they hardcoded the same dead slugs separately).
3. Don't mark a model "supported"/"verified" until a real end-to-end completion has run — passing `LlmModel.valid()` only proves the string is accepted, not that the provider serves it.

Refs: #851, #846.

samin1554:
Hi, I'd like to take a stab at this. I may have a solution as I was exploring in my local machine.

andreakiro:
@samin1554 Hey, thanks for checking in. Is your stuff ready for review?

PR #855, opened 2026-06-27
feat(core): add Fireworks provider with prefix-based validation

samin1554:
**Summary**
Fireworks models rotate pretty frequently and the current approach of hardcoding each one as a LlmModel enum entry means every rotation needs an SDK change and a release. This PR fixes that by validating at the provider level instead.
**What changed**

- packages/notte-core/src/notte_core/common/config.py
- Added LlmProvider.fireworks_ai with FIREWORKS_API_KEY mapping
- Added is_prefix_provider property on LlmProvider to flag providers that accept arbitrary model paths
- Added LlmModel.is_valid(model) which handles both enum models and prefix providers. Cloud can migrate from model in LlmModel.valid() to LlmModel.is_valid(model)
- Added fireworks_ai to the use_strict_response_format unsupported list
- No new LlmModel enum entries for Fireworks, that's kind of the point
- tests/config/test_fireworks_provider.py

15 tests covering provider registration, arbitrary path validation, API key gating, response format, and making sure non-prefix providers still reject unknown models

**How it addresses #854**

Any fireworks_ai/* path is accepted when FIREWORKS_API_KEY is set, no SDK release needed for model rotations
No hardcoded slugs in the enum so docs and SDK can't drift independently
is_valid() confirms the string is structurally valid and the key exists, whether the model is actually deployed is on the provider at runtime

Migration note
valid() is unchanged for backwards compat. The cloud endpoint just needs to swap model in LlmModel.valid() for LlmModel.is_valid(model).

## Summary by CodeRabbit

* **New Features**
  * Added Fireworks AI as a supported LLM provider, including environment-based API key handling and wider model path matching.
  * Expanded model validation to accept Fireworks AI–prefixed models when the required API key is present.
* **Bug Fixes**
  * Applied correct response-format handling and stricter validation for unsupported/unknown model names.
  * Improved request validation for “prefix providers” to ensure models are only accepted when their environment requirements are satisfied.
* **Tests**
  * Added coverage for Fireworks AI provider behavior, including edge cases and environment-variable scenarios.

### DIFF
--- a/packages/notte-core/src/notte_core/common/config.py
+++ b/packages/notte-core/src/notte_core/common/config.py
@@ -92,7 +92,7 @@ class LlmProvider(StrEnum):
 
     @property
     def apikey_name(self) -> str:
-        if enable_openrouter():
+        if enable_openrouter() and self != LlmProvider.fireworks_ai:
             return "OPENROUTER_API_KEY"
         match self:
             case LlmProvider.gemini:
@@ -265,7 +265,8 @@ class LlmModel(StrEnum):
             return True
         try:
             provider = LlmModel.get_provider(model)
-            return provider.is_prefix_provider and provider.has_apikey_in_env()
+            _, _, suffix = model.partition("/")
+            return provider.is_prefix_provider and bool(suffix) and provider.has_apikey_in_env()
         except ValueError:
             return False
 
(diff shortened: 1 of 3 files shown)


leo-notte:
Thanks for taking this on and for implementing the provider-level validation proposed in #854. For now, we're standardizing access to these model families through our existing OpenRouter integration (including Kimi, GLM, and MiniMax) rather than adding and maintaining a separate direct Fireworks provider path with its own credentials and validation. Closing this PR as a product-scope decision, not because of the quality of the contribution.

PR #887, opened 2026-08-06
Fix blob downloads and flaky new-tab observation

giordano-lucas:
Two integration failures, two independent causes.

## Blob downloads never worked

`download_file` on a button that builds its payload with `URL.createObjectURL` always failed with `FailedToDownloadFileError`.

The controller installed a capture hook that wraps `URL.createObjectURL`, then read the Blob back by URL after the click. Both ran through `page.evaluate`, which executes in an **isolated world**. The page's own script calls the real `URL.createObjectURL` in the main world, so the hook never saw a single call and the lookup map stayed empty:

```
DEBUG dl_url='blob:https://.../86cbf81a-...'
DEBUG hook_present=True
DEBUG keys=[]          # <- nothing ever captured
DEBUG b64=None         # -> FailedToDownloadFileError
```

Confirmed directly: a global set by the page's inline script is invisible to `page.evaluate`, and vice versa.

Both snippets now run over CDP `Runtime.evaluate`, which targets the main world. That also means they are not subject to the page's CSP, unlike an injected `<script>` tag.

Verified against all five download fixtures, byte-for-byte:

```
anchor_download_attr: OK (23 bytes)
blob_button:          OK (21 bytes)     <- was failing
raw_txt:              OK (23 bytes)
raw_jpg:              OK (1970107 bytes)
raw_pdf:              OK (223121 bytes)
```

## `test_observe_with_instructions` was racing

Every console link on the landing page carries `target="_blank"`, so the click always lands in a new tab. The test's `switch_tab` ran *after* `aobserve()`, so it could not affect the assertion - the observation was taken on whichever tab happened to be current. It passed only when the controller's own new-tab detection (which polls the page count after a short wait) won the race, and failed with `Expected URL to start with https://console.notte.cc, got https://www.notte.cc/` when it lost.

Moved the switch before the observation.

---
<a href="https://app.blacksmith.sh/nottelabs/codesmith/notte/pr/887"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788596797&installation_model_id=8247&pr_number=887&repository=nottelabs%2Fnotte&return_to=https%3A%2F%2Fgithub.com%2Fnottelabs%2Fnotte%2Fpull%2F887&signature=75e795b365f144721303ff2355ed69bc3e54cda2e0bda71041a96fe0bf12f4bf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blob-based download handling, including files created and revoked by page scripts.
  * Increased download reliability with improved cleanup after each download.
  * Preserved support for authenticated non-blob downloads across supported browsers.
  * Improved browser compatibility and error handling during download capture.

* **Tests**
  * Updated browser interaction coverage for links that open in new tabs.
  * Added coverage for download evaluation, error handling, cross-origin blobs, and cleanup.

### DIFF
--- a/packages/notte-browser/src/notte_browser/controller.py
+++ b/packages/notte-browser/src/notte_browser/controller.py
@@ -59,7 +59,7 @@ from notte_browser.errors import (
     capture_playwright_errors,
 )
 from notte_browser.form_filling import FormFiller
-from notte_browser.playwright_async_api import CDPSession, Locator
+from notte_browser.playwright_async_api import CDPSession, Frame, Locator, evaluate_in_main_world
 from notte_browser.window import BrowserWindow
 
 # Installed once per download action. Wraps URL.createObjectURL so we retain a
@@ -68,10 +68,9 @@ from notte_browser.window import BrowserWindow
 # Revocation only removes the URL->Blob mapping in the browser's registry;
 # our Map keeps the Blob object alive.
 #
-# Both snippets must run in the page's main world: the Blob is created by the
-# page's own script, and `page.evaluate` runs in an isolated world, where the
-# hook would patch a different `URL` and never observe the page's calls. Driving
-# them over CDP puts them in the main world and also sidesteps the page CSP.
+# Both snippets must run in the page's main world, where the Blob is created.
+# Chromium's top frame uses CDP; iframe and Firefox evaluation explicitly uses
+# the owning frame's page world rather than Patchright's default isolated world.
 _BLOB_CAPTURE_HOOK = """
 (() => {
   if (window.__notte_blob_capture) return;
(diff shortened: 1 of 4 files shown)


eeshsaxena (commented):
The blob path fix makes sense. The core issue is that the capture hook and the read script have to share `window.__notte_blob_capture`, and running both through a single main-world `Runtime.evaluate` on one dedicated CDP session guarantees they land in the same execution context, whereas splitting them across separate evaluate calls risked landing in different worlds and getting a null read. Inlining the url via `json.dumps` inside the IIFE is also safe against quoting/injection. The `try/finally: await cdp.detach()` cleans the session up, and since `cdp` is assigned before the `try`, a failure in `new_cdp_session` itself won't reach an unbound `detach`.

One small thing: `_main_world_evaluate` doesn't inspect `exceptionDetails` on the CDP response, so if the read script throws in-page it comes back as a non-string `value` and surfaces as `FailedToDownloadFileError` rather than the underlying JS error. Fine for the failure path, it just means the real cause is invisible in logs if this ever regresses.

leo-notte:
@greptileai review

leo-notte:
@greptileai review

leo-notte:
@greptileai review

PR #904, opened 2026-08-24
fix(core): read collection variables sent as strings

giordano-lucas:
The runtime half of a bug already patched twice in frontends. This fixes it once, for every caller.

## It does not crash. It returns garbage.

A caller sends `keywords="['a', 'b']"` for `keywords: list[str]`. The script receives a 21-character string, and nothing raises:

```python
def run(keywords: list[str]):
    return {"count": len(keywords), "joined": ",".join(keywords)}
```

| | result |
|---|---|
| **before** | `{'count': 21, 'joined': "[,',a,',,, ,',b,',]"}` |
| **after** | `{'count': 2, 'joined': 'a,b'}` |

`join` walks the string character by character. The function completes, returns, and is recorded as a successful run. That is worse than a traceback, because it looks like an answer.

## Why every client hits it

`ParameterInfo.default` holds Python *source* — it is what `ast.unparse` wrote when the parameter was recorded. Any caller that pre-fills a payload from a stored default is echoing that source back, and `JSON.parse` rejects its single quotes. The console and the Anything marketplace playground were each patched separately for exactly this; the SDK, CLI, MCP `run` and direct API callers still have it.

## What changed

Collection-typed variables arriving as strings are parsed with `json.loads`, then `ast.literal_eval`.

**Literals only.** A name like `DEFAULT_KEYWORDS` or a call like `list()` raises, and the value passes through untouched. `set()` is CPython's one exception — special-cased because there is no empty-set literal — so a parameter defaulted that way arrives as `[]`, which is the value it asked for rather than four characters of source. There is a test pinning that, because it surprised me.

**The outermost annotation decides.** Read with `ast.parse` rather than pattern-matched, since `ast.unparse` produced these strings and `ast.parse` reads them back exactly. So `dict[str, list[str]]` is a mapping, not a sequence that happens to mention one — handing a dict parameter a list would be the same silently wrong argument in the other direction. `Optional[...]`, `Annotated[...]`, `Union[...]` and `X | None` are looked through; a parameter that has a default is very often wrapped.

## Why it is this narrow

Checked against production before touching anything:

- **`int` and `bool` parameters receiving strings succeeded 5,166 times in a fortnight.** Python is duck-typed and the scripts cope. Coercing or rejecting those would break working calls, so they are untouched.
- **A collection parameter receiving a string failed 4 times out of 4 over thirty days.** A 100% failure rate. There is no behaviour to preserve and nothing this can regress.

That asymmetry is the entire justification for the scope. Nothing is rejected, nothing else is converted, and a value this cannot read is passed through exactly as it arrived.

## One structural note

The coercion runs inside `_validate_variables` and mutates in place. That is deliberate and worth flagging in review: the workflows Lambda replaces `run_script` wholesale and calls this method on the same dict it later splats into `run(**variables)`. A separate coercion step would be tidier and would miss the path that actually runs published functions.

## Verification

26 new tests. `ruff check` introduces zero new errors against the file's pre-existing 9; `basedpyright` reports 6 errors / 5 warnings both before and after, none in the new code; `ruff format` clean.

End-to-end through `SecureScriptRunner.run_script` with a real script: the Python-literal spelling and the JSON spelling both arrive as a list, an already-structured value is untouched, and `limit="20"` on an `int` parameter stays a `str`.

---
<a href="https://app.blacksmith.sh/nottelabs/codesmith/notte/pr/904"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790177406&installation_model_id=8247&pr_number=904&repository=nottelabs%2Fnotte&return_to=https%3A%2F%2Fgithub.com%2Fnottelabs%2Fnotte%2Fpull%2F904&signature=290226a946c9bbafb91e3bee3f8757ec365102fe301b99c58b698b7e0d7fda84"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Automatically converts stringified JSON and Python-literal values into supported sequence and mapping variables.
  - Supports nested collections, unions, common collection annotations, and quoted annotations such as `"list[str]"`.
  - Preserves already structured values and safely ignores invalid, unsupported, or non-collection inputs.
  - Normalizes collection values before execution for more consistent parameter handling.
  - Safely handles unusually large or deeply nested inputs without interrupting execution.

- **Tests**
  - Added coverage for collection formats, nested values, quoted annotations, empty collections, invalid expressions, parser failures, and missing variables.

### DIFF
--- a/packages/notte-core/src/notte_core/ast.py
+++ b/packages/notte-core/src/notte_core/ast.py
@@ -67,6 +67,9 @@ def _container_kind(annotation: str | None) -> str | None:
     a mapping, since guessing a sequence for something that might be a dict is
     the more expensive mistake.
 
+    Quoted annotations are unwrapped and re-read, since `ast.unparse` keeps the
+    quotes a forward reference was written with.
+
     Parsed rather than pattern-matched: `ast.unparse` produced these strings, so
     `ast.parse` reads them back exactly, including nesting the eye slides over.
     """
@@ -81,6 +84,17 @@ def _container_kind(annotation: str | None) -> str | None:
 
     def walk(current: ast.expr, depth: int = 0) -> None:
         if depth > 8:  # A pathological annotation is not worth recursing into.
+            return
+        if isinstance(current, ast.Constant) and isinstance(current.value, str):
+            # A quoted annotation. `def f(x: "list[str]")` is recorded by
+            # `ast.unparse` as the six characters `'list[str]'`, quotes and all,
+            # so without this the string reads as a bare constant of no
+            # particular type and the parameter silently keeps its raw value.
+            # Reached again for the inner half of `Optional["list[str]"]`.
+            try:
+                walk(ast.parse(current.value.strip(), mode="eval").body, depth + 1)
+            except SyntaxError:
+                pass
             return
         if isinstance(current, ast.BinOp) and isinstance(current.op, ast.BitOr):
             walk(current.left, depth + 1)
@@ -139,7 +153,13 @@ def _parse_collection_text(text: str, kind: str) -> Any | None:
     """
     try:
         value: Any = json.loads(text)
-    except ValueError:
+    except (ValueError, RecursionError):
+        # RecursionError alongside ValueError because `json.loads` recurses per
+        # nesting level: deeply nested input raises it rather than rejecting the
+        # text, and letting it escape would end the run over a variable this
+        # function is allowed to decline. Declining means passing the original
+        # string through, which is what happens if both parsers fail.
+        #
         # Nested rather than a loop with `continue`, so neither attempt needs a
         # bare `except` and the exceptions each parser really raises stay named.
         try:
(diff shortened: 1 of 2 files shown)


giordano-lucas:
@greptile review

giordano-lucas:
@greptile review

PR #906, opened 2026-08-24
feat(core): bring the script runner up to the runtime version

giordano-lucas:
`notte_core.ast` and the copy the hosted runtime executes had drifted apart.
This closes the gap in one direction, because the drift is one directional:

| | notte_core (before) | runtime copy |
|---|---|---|
| lines | 856 | 1466 |
| top level symbols | 13 | 21 |

Every symbol `notte_core` had, the runtime copy also had. **Nothing here was
ahead of it**, so there was nothing to merge, only to catch up. Of the 13 shared
symbols, 10 were already identical; the 3 that differed were
`MissingRunFunctionError`, `ScriptValidator` and `SecureScriptRunner`.

### Behaviour that changes

**Builtins are now restricted in the default mode.** `run_script` defaults to
`restricted=False`, and that path previously set no `__builtins__` key at all,
which makes CPython inject the real builtins. Scripts therefore had `eval`,
`exec`, `compile`, `open` and `__import__`. They now get an allowlist of 70
names, with `__import__` and `open` swapped for checked versions.

**`open` is permitted, and confined to /tmp.** It used to be rejected by the
validator at parse time. It now resolves to a wrapper that resolves the path
first, then refuses anything that is not under /tmp, so `..` and symlinks cannot
walk out, and relative names land in /tmp rather than the working directory.

Net: considerably stricter on builtins, deliberately looser on scratch files.
The first is the larger change, and the one worth reading closely if you embed
this to run untrusted code.

**Run entry point resolution** also arrives: several candidate `run` definitions,
a rebinding, or a `__name__ == "__main__"` guard are now diagnosed rather than
resolved by accident.

### Tests

`tests/scripts`: **79 passed**, 5 skipped. The one failure on this branch,
`test_run_script_with_agent`, fails identically on `main` and is an absent
Gemini API key, not this change.

Two existing tests changed:

- `test_private_attribute_forbidden` - still refused, wording moved from
  "private attribute" to "dangerous attribute". `__class__`, `__bases__` and
  `__subclasses__` are all still blocked; verified by hand.
- `test_dangerous_builtins_forbidden` - the `open` half now parses, so it
  asserts on `eval` instead, and the confinement is covered by a new test.

`test_open_is_confined_to_tmp` is new and pins the containment directly,
including the `/tmp/../etc/passwd` walk out.

`ruff` and `basedpyright` are clean (3 pre-existing warnings, unchanged).

---
<a href="https://app.blacksmith.sh/nottelabs/codesmith/notte/pr/906"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790184851&installation_model_id=8247&pr_number=906&repository=nottelabs%2Fnotte&return_to=https%3A%2F%2Fgithub.com%2Fnottelabs%2Fnotte%2Fpull%2F906&signature=e23dce639a2e369835119c0f394ba8129a6d1655ea304e201c2748133a3948aa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering both synchronous and asynchronous `run` entry points.
  * Added collection-type input handling and improved script parsing.
  * Added safe temporary-file access within `/tmp`.

* **Bug Fixes**
  * Prevented ambiguous, conditional, decorated, or overridden `run` definitions from executing unexpectedly.
  * Improved sandbox protection against unsafe imports, attributes, built-ins, and file paths.
  * Preserved existing logger handlers during script execution.

### DIFF
--- a/packages/notte-core/src/notte_core/ast.py
+++ b/packages/notte-core/src/notte_core/ast.py
@@ -1,34 +1,218 @@
-"""Script parsing and sandboxed execution for Notte functions.
+"""Parsing and sandboxed execution of user-authored Notte functions.
 
-The variable coercion helpers here (`_annotation_head`, `_container_kind`,
-`_parse_collection_text`, `coerce_collection_variables`, the annotation head
-frozensets they share, and `_validate_variables`) are mirrored verbatim by the
-Notte API runtime, which keeps its own fork of this module. Keep the two copies
-byte-identical.
+`ScriptValidator` decides what a submitted script is allowed to contain and
+pulls the signature of its `run` entry point back out. `SecureScriptRunner`
+executes it, either under RestrictedPython or, by default, in plain Python with
+a builtins allowlist that withholds `eval`, `exec`, `compile`, `open` and
+`__import__`, substituting sandboxed versions of the last two.
 
-A fix applied to only one of them leaves the other passing a JSON string
-straight through to a parameter annotated `list` or `dict`. That raises nothing:
-a function expecting `list[str]` and handed '["a", "b"]' iterates 21 characters
-and returns a wrong answer instead of failing.
+Variables arriving from a caller are coerced to match the entry point's
+annotations before it is invoked, so a JSON or Python-literal string sent for a
+`list` or `dict` parameter becomes the collection it spells out rather than
+being iterated character by character.
 """
 
 import ast
+import builtins
+import copy
 import json
+import re
+import sys
 import traceback
 import types
-from collections.abc import Iterable, Mapping
-from typing import Any, Callable, ClassVar, Literal, Protocol, cast, final
+from collections.abc import Iterable
+from pathlib import Path
+from typing import Any, ClassVar, Literal, Protocol, cast, final
 
 from pydantic import BaseModel, ConfigDict
-from RestrictedPython import compile_restricted, safe_globals  # type: ignore [reportMissingTypeStubs]
-from RestrictedPython.transformer import RestrictingNodeTransformer  # type: ignore [reportMissingTypeStubs]
+from RestrictedPython import (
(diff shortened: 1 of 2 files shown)


giordano-lucas:
Closing in favour of targeted changes.

This replaced the file wholesale, which bundles three different kinds of change into one commit: security hardening the package genuinely lacks, deliberate loosenings that are right for one runtime but not obviously right for a general sandbox, and unrelated features. That is not reviewable as a unit, and the loosenings would have ridden in alongside the fixes.

The hardening worth having on its own merits, in priority order:

1. `DISALLOWED_IMPORT_MEMBERS`. `asyncio` is on the import allowlist and nothing guards its members, so `asyncio.create_subprocess_shell`, `to_thread` and `subprocess` are all reachable from a script today.
2. A builtins allowlist for the default unrestricted mode, which currently sets no `__builtins__` key and so inherits the real ones, including `eval`, `exec`, `compile`, `open` and `__import__`.
3. `safe_getattr` defaults to `None`, so a missing attribute is returned rather than raised.

Will reopen as separate PRs.

PR #912, opened 2026-08-25
fix(core): deny subprocess creation while a function runs

giordano-lucas:
## The hole

`SecureScriptRunner.run_script` runs a function in **unrestricted** mode by
default - a plain `exec` whose `__import__` refuses `os`, `subprocess`,
`socket`, `ctypes` and the rest, with a builtins allow list that withholds
`eval`/`exec`/`open`. `asyncio` is deliberately allowed, because functions use
`asyncio.to_thread` / `run_in_executor` for concurrency, and asyncio is the leak:

```python
run_script(restricted=False) on:
    import asyncio
    async def go():
        p = await asyncio.create_subprocess_shell("echo PWNED-$(id -u)", stdout=-1)
        ...
# -> 'PWNED-501'
```

Anyone running `SecureScriptRunner` on untrusted code in unrestricted mode is
exposed. In this SDK that is the `Function.run(local=True, restricted=False)`
path, which executes downloaded function code in the caller's own interpreter.
(`restricted=True`, the default, blocks the async form, so the default local run
is not exposed - but `restricted=False` is one argument away.)

## The fix

`deny_subprocess_creation()` wraps the user-code `exec` in both branches.
Blocking the two `create_subprocess_*` names is **not** enough - I verified the
bypass: they, the event loop's `subprocess_shell`/`subprocess_exec`, and the
low-level `loop._make_subprocess_transport` all funnel through
**`subprocess.Popen`** on Linux, so that is the choke point. Neutering it costs
nothing legitimate because `import subprocess` is already refused.

**Scoped, not a global import-time patch** - deliberately. A global patch breaks
two things: a runtime that spawns its own worker children through a forkserver
(the patch would deny that too), and a host process that must stay able to spawn
after a local run. The guard saves originals and restores on exit; a test
asserts the restore.

## Limits

Save-and-restore is not reentrant across threads sharing one process; the runner
executes one function per child, so that does not arise. This is a process-level
mitigation - a hostile guest still warrants an OS boundary (seccomp denying
`fork`/`execve`), which this does not replace.

## Tests

`tests/scripts/test_subprocess_denied.py`: the shell escape and the
`_make_subprocess_transport` bypass both raise; `asyncio.to_thread` still works;
`subprocess.Popen` is restored after the guarded block. Full `tests/scripts`
suite green apart from the pre-existing `test_run_script_with_agent` (absent
Gemini key, fails identically on `main`). `ruff` and `basedpyright` clean.

---
<a href="https://app.blacksmith.sh/nottelabs/codesmith/notte/pr/912"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790271582&installation_model_id=8247&pr_number=912&repository=nottelabs%2Fnotte&return_to=https%3A%2F%2Fgithub.com%2Fnottelabs%2Fnotte%2Fpull%2F912&signature=e9d51a78cd41f583ea97f8e4618b349395d4b74491278c4fa0d6dcfcc75d34d9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Security**
  * Script execution now blocks attempts to create subprocesses, including through asynchronous process APIs.
  * Subprocess restrictions remain enforced during nested and concurrent executions.
  * Existing process-spawning behavior is restored after restricted execution ends.

* **Bug Fixes**
  * Improved enforcement across supported execution paths.
  * Confirmed that non-process background tasks continue to work as expected.

### DIFF
--- a/packages/notte-core/src/notte_core/ast.py
+++ b/packages/notte-core/src/notte_core/ast.py
@@ -17,6 +17,7 @@ import contextlib
 import json
 import os
 import subprocess
+import threading
 import traceback
 import types
 from collections.abc import Iterable, Iterator, Mapping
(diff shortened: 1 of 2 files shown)


giordano-lucas:
@greptile review

giordano-lucas:
Both P1s addressed. The guard is now reentrant (depth count under a lock): overlapping local runs share one patch and restore only when the last finishes, so no corruption and no early-restore gap — tests cover nested and cross-thread overlap. On the background-thread escape: that gap is not closable for an in-process restoring guard, and the docstring now says so plainly — restricted mode (the default) refuses the async def that reaches asyncio's subprocess API, and the Lambda runner (nottelabs/monorepo#2336) closes it properly with a single-use child + permanent denial. This is a defense-in-depth layer, not the boundary. @greptile review

giordano-lucas:
Closing. The review found the root problem this approach can't solve: a process-global subprocess/os patch in `run_script` blocks the **host's own** subprocess calls, not just the guest's. `Function.run(local=True)` runs in the caller's process, and module attributes are process-global (not thread-local), so during any local run a host thread's legitimate `subprocess.Popen` also raises `SubprocessNotAllowed`. A library must not break its caller's process capabilities as a side effect, and there's no in-process way to distinguish guest calls from host calls.

So the in-process guard is the wrong tool here. `restricted=True` (the default) already refuses the `async def` that reaches asyncio's subprocess API; `restricted=False` is unsandboxed by design. The untrusted-execution surface that genuinely needs a subprocess boundary is the hosted runner, which runs each function in a dedicated single-use child — no host to interfere with — and denies subprocess there permanently. That fix lives in nottelabs/monorepo#2336 (accepted at 4/5).

The durable boundary for hostile code remains an OS-level one (seccomp denying fork/execve), tracked separately.

PR #922, opened 2026-08-28
ci: merge the automation prs from the job that opens them

andreakiro:
The two scheduled workflows (`bump-submodules` and `refresh-llms`) open a pull request and call `gh pr merge --auto`, but those pull requests never merge on their own.

`main` requires one approving review. A bypass allowance lets a permitted user merge without that review, but GitHub does not extend the bypass to the auto-merge queue acting on their behalf, so the pull request stays at `REVIEW_REQUIRED` indefinitely even with every check green. Performing the merge explicitly with the same token does honour the bypass.

This replaces the `--auto` call in both workflows with an explicit merge: the job waits for the pull request's checks, then merges it.

### Why the merge lives in the job that opens the pull request

The first version of this change was a separate workflow triggered on `pytest cicd` completion, which located the pull request by head branch name. That is unsafe. A fork pull request can use the same branch name, and `pytest cicd` runs on fork pull requests and succeeds quickly because the secret-backed steps are skipped for untrusted heads. The lookup would then have matched the fork's pull request and merged it into `main` with a token that bypasses review.

Merging from the job that opened the pull request removes that class of problem rather than filtering for it. The job already holds the pull request number it just created on a branch it pushed itself, so nothing is resolved by name and there is nothing for an outside branch to impersonate.

### Behaviour

- Waits for the checks to be reported, then watches them, failing fast on the first failure.
- Merges with squash and deletes the branch when they pass.
- When they do not pass, the run ends with a warning and leaves the pull request open. The next scheduled run rebuilds the branch from `main` and retries, so a flaky check costs a day rather than wedging the automation.
- The job timeout rises to 45 minutes to cover the wait.

## Summary by CodeRabbit

* **Improvements**
  * Automated maintenance pull requests now wait for checks to be reported and complete before merging.
  * Successful checks trigger a squash merge and branch cleanup.
  * Failed checks leave the pull request open for retry during the next scheduled run.
  * Workflow time limits were extended to better accommodate check and merge processing.


andreakiro:
Issue 1 is correct and the fix was to remove the separate workflow entirely rather than add trust checks to it.

The lookup was by head branch name, and `pytest cicd` does run on fork pull requests, succeeding quickly because `IS_TRUSTED` is false and the secret-backed steps are skipped. A fork branch named `chore/bump-submodules` would therefore have been selected and merged into `main` with the review-bypassing token.

The merge now happens inside the job that opens the pull request, using the number it just created on a branch it pushed itself. No pull request is resolved by name, so the impersonation path does not exist. `.github/workflows/automerge-bot-prs.yml` is deleted.

On issue 2: there is no runner for GitHub Actions workflows in this repository, so the behaviour was instead verified directly. The guard logic in `refresh-llms` was exercised against the live OpenAPI spec in both the healthy and unreachable cases, and `bump-submodules` was run end to end via `workflow_dispatch`, which fast-forwarded all five submodules and opened #921. Both workflows remain `workflow_dispatch`-enabled so they can be exercised manually.

@greptile review

